PR #3, opened 2020-05-27
[UCA-432] added extended timeout when user is providing custom script

omazapa:
tested sourcing a script with sleep 30 

thanks to @prasanthkothuri  for helping me with the spare node :)

### DIFF
--- a/SwanSpawner/swanspawner/swanspawner.py
+++ b/SwanSpawner/swanspawner/swanspawner.py
@@ -11,7 +11,8 @@ from socket import gethostname
 from traitlets import (
     Unicode,
     Bool,
-    Dict
+    Dict,
+    Int
 )
 
 from tornado import gen
@@ -69,9 +70,11 @@ def define_SwanSpawner_from(base_class):
             help='Extra environment variables to pass to the container',
         )
 
-        default_timeout = 60
-
-        extended_timeout = 120
+        extended_timeout = Int(
+            default_value=120,
+            config=True,
+            help="Extended timeout for users using environment script"
+        )
 
         def __init__(self, **kwargs):
             super().__init__(**kwargs)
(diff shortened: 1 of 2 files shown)


SwanSpawner/swanspawner/swanspawner.py
```diff
@@ -69,6 +69,10 @@ class SwanSpawner(base_class):
             help='Extra environment variables to pass to the container',
         )
 
+        default_timeout = 60
```
diocas, line 72: Please make these configurable

jupyterhub_config.py
```diff
@@ -26,6 +26,6 @@
 c.SwanSpawner.http_timeout = 30
 
 # Spawner.start_timeout - https://jupyterhub.readthedocs.io/en/stable/api/spawner.html#jupyterhub.spawner.Spawner.start_timeout
-c.SwanSpawner.start_timeout = 60
+#c.SwanSpawner.start_timeout = 60
```
diocas, line 29: We should keep this config and just add another one for the extended.

SwanSpawner/swanspawner/swanspawner.py
```diff
@@ -178,6 +182,12 @@ def start(self):
             """
 
             start_time_start_container = time.time()
+            
+            #if the user script exists, we allow extended timeout
+            if self.user_options[self.user_script_env_field].strip()=='':
+                self.start_timeout = self.extended_timeout
+            else:
```
diocas, line 189: please remove the else and keep the default behavior unchanged.

SwanSpawner/swanspawner/swanspawner.py
```diff
@@ -186,8 +189,6 @@ def start(self):
             #if the user script exists, we allow extended timeout
             if self.user_options[self.user_script_env_field].strip()=='':
```
diocas, line 190: If the script field is empty, you use then you use the extended timeout? Am I missing anything?

PR #4, opened 2020-05-28
Binder integration to SWAN

prasanthkothuri:


SwanSpawner/swanspawner/swandockerspawner.py
```diff
@@ -150,17 +156,18 @@ def start(self):
...
         """
 
+        # default values when spawned via API (e.g binder)
         username = self.user.name
-        platform = self.user_options[self.platform_field]
-        lcg_rel = self.user_options[self.lcg_rel_field]
-        cluster = self.user_options[self.spark_cluster_field]
-        cpu_quota = self.user_options[self.user_n_cores]
-        mem_limit = self.user_options[self.user_memory]
+        platform = self.user_options.get(self.platform_field,'x86_64-centos7-gcc8-opt')
```
diocas, line 161: Can we take the default values from the default configuration in the json file?
Then we no longer need to keep this up to date with valid values.

SwanSpawner/swanspawner/swandockerspawner.py
```diff
@@ -105,6 +105,12 @@ def get_env(self):
             self.extra_host_config['port_bindings'] = {}
             self.extra_create_kwargs['ports'] = []
 
+            if self.lcg_rel_field not in self.user_options:
+                # session spawned via the API, in binder start notebook with jovyan user
+                self.extra_create_kwargs['working_dir'] = "/home/jovyan"
+                self.extra_create_kwargs['user'] = 'jovyan'
+                self.extra_create_kwargs['command'] = ["jupyterhub-singleuser","--ip=0.0.0.0","--NotebookApp.default_url=/lab"]
```
diocas, line 112: Is this done in the upstream spawner? Isn't this mandatory as the container entry point?

KeyCloakAuthenticator/keycloakauthenticator/auth.py
```diff
@@ -72,7 +72,8 @@ async def authenticate(self, handler, data=None):
             return None
 
     async def pre_spawn_start(self, user, spawner):
-        spawner.user_roles = self.user_roles
+        if hasattr(self, 'user_roles'):
```
diocas, line 75: This is not related to binder integration but is in the same commit.

SwanSpawner/swanspawner/swanspawner.py
```diff
@@ -94,29 +94,41 @@ def get_env(self):
...
+                    ROOT_LCG_VIEW_PLATFORM = self.user_options[self.platform_field],
+                    USER_ENV_SCRIPT        = self.user_options[self.user_script_env_field],
+                    ROOT_LCG_VIEW_PATH     = self.lcg_view_path,
+                    USER                   = username,
+                    USER_ID                = str(userid),
+                    HOME                   = homepath,
+                    EOS_PATH_FORMAT        = self.eos_path_format,
+                    SERVER_HOSTNAME        = os.uname().nodename,
+
+                    JPY_USER               = self.user.name,
```
diocas, line 116: We probably don't need this, but let's clean it in another PR.

prasanthkothuri: yes I noticed that these are not required, we can clean up in a separate PR

KeyCloakAuthenticator/keycloakauthenticator/auth.py
prasanthkothuri: it is related as when we launch with binder auth object is not formed on jupyterhub

SwanSpawner/swanspawner/swandockerspawner.py
prasanthkothuri: good point, I will change it

SwanSpawner/swanspawner/swandockerspawner.py
prasanthkothuri: maybe in kubespawner but for systemuserspawner it doesn't work without these, working_dir was defaulting to /home/{user.name} and command to jupyter notebook --ip=0.0.0.0

SwanSpawner/swanspawner/swandockerspawner.py
prasanthkothuri: thinking about these values, they don't have to be latest, they just need to be a valid combination, we can leave that as static default values

diocas: I do not agree. Memory and cores, for example, are always used right? So they should reflect the default values.
And this, which is a very simple addition to the code, might avoid any problems in the future with sciencebox.

prasanthkothuri: ok, done now, taking the values from the config json

KeyCloakAuthenticator/keycloakauthenticator/auth.py
diocas: The user auth state? But shouldn't we still have roles? Or when you start with Binder, the user not the real user?
And does it matter if you start from binder or hub? Isn't the auth always happening in Hub?

SwanSpawner/swanspawner/swandockerspawner.py
```diff
@@ -149,18 +156,22 @@ def start(self):
...
+        with open(self.options_form_config) as json_file:
+            options_form_config_data = json.load(json_file)
 
         username = self.user.name
-        platform = self.user_options[self.platform_field]
-        lcg_rel = self.user_options[self.lcg_rel_field]
-        cluster = self.user_options[self.spark_cluster_field]
-        cpu_quota = self.user_options[self.user_n_cores]
-        mem_limit = self.user_options[self.user_memory]
+        platform = self.user_options.get(self.platform_field,options_form_config_data["options"][1]['platforms'][0]['value'])
```
diocas, line 165: I would prefer having a function like `get_default_config()`, store the output in a variable, and then use it here. But I don't mind that much.

prasanthkothuri: thought about it but since we use these only in one place I did it like this

KeyCloakAuthenticator/keycloakauthenticator/auth.py
diocas: Binder impersonates the users and is not the user, so it doesn't have its state.

PR #14, opened 2020-07-06
[UCA-443] removed support to use swap memory, setting swappiness to zero

omazapa:
According to 
https://docs.docker.com/config/containers/resource_constraints/#--memory-swappiness-details

swappiness should be zero.

using docker inspect to the container  in swan-spare003,
I can see that the value changed from 

"MemorySwappiness": null,
to
"MemorySwappiness": 0

Cheers
Omar.

### DIFF
--- a/SwanSpawner/swanspawner/swandockerspawner.py
+++ b/SwanSpawner/swanspawner/swandockerspawner.py
@@ -106,6 +106,9 @@ class SwanDockerSpawner(define_SwanSpawner_from(SystemUserSpawner)):
             self.extra_host_config['port_bindings'] = {}
             self.extra_create_kwargs['ports'] = []
 
+            #disabling swap memory usage
+            self.extra_host_config['mem_swappiness'] = 0
+
             if self.lcg_rel_field not in self.user_options:
                 # session spawned via the API, in binder start notebook with jovyan user
                 self.extra_create_kwargs['working_dir'] = "/home/jovyan"


diocas:
@omazapa Was this change in the docker configuration changed recently? Why did we only notice now? (maybe because we updated the docker version a few months ago?)
Is there any configuration for this in jupyterhub, besides setting this in the spawner? I would prefer not to add things to the spawner.

omazapa:
@diocas I am not sure, but If we changed the docker version recently, that can be a good reason, because in our spawner we are leaving this parameter by default, and maybe with the new version docker is allowing use swap.

 The swappiness have 3 options in docker

--memory-swappiness details (https://docs.docker.com/config/containers/resource_constraints/#--memory-swappiness-details)

*  A value of 0 turns off anonymous page swapping.
*  A value of 100 sets all anonymous pages as swappable.
*  By default, if you do not set --memory-swappiness, **the value is inherited from the host machine**. (this means null)

I checked the production nodes with docker inspect, I can see that the containers have by default 
"MemorySwappiness": null,

With this option set to zero, the container can not be able to use any swan for nothing.


Another important thing is that almost all the swan memory is used by processes like eosd and cvmfs2,
you can take a look with the command 
`
for file in /proc/*/status ; do awk '/VmSwap|Name/{printf $2 " " $3}END{ print ""}' $file; done | sort -k 2 -n -r | less
`

let me know what do you think about this approach or if we need further research about this.

diocas:
@omazapa Thanks, that information is fine. But I was also asking if JupyterHub has a configuration for this option. I would prefer to do it in the configuration than adding an extra line of code.

omazapa:
closing this because I did the changes in puppet.

PR #15, opened 2020-07-09
Authenticator updates

### LINKED ISSUE #11: Remove dependency in PWD
This functionality should come from the authenticator. This is fundamental to any authentication that is not local/ldap (oauth for example).

https://github.com/swan-cern/jupyterhub-extensions/blob/b1ad71800cf09a8e9a24404253775bba884025a8/SwanSpawner/swanspawner/swanspawner.py#L111

### LINKED ISSUE #13: Update oAuth user information
To update the oAuth tokens that we need for swan-cern/jupyter-extensions#27, it would help if we always had the latest tokens in JupyterHub. Since the upstream authenticator hasn't implemented `refresh_user`, I suggest we do it in our own.

This code looks like it's what we want and is compatible with keycloak: https://github.com/jupyterhub/oauthenticator/pull/287

diocas:
Changes in this PR related to the authenticator:
* Easy configuration by using the .wellknown to fetch all the necessary endpoints
* Updated readme and installation (to push this to pypi)
* Automatic refresh of the oAuth tokens
* Configurable way of retrieving the uid of the user and cleaner way of passing it to the spawner
* Access token passed as environment variable to be used in the startup script

Closes #13 
Closes #11


diocas:
@ricastell This means that the first part of the token update is done. Now you will have a token set as env variable, so in the session startup script you will need to save it to the corresponding file so that EOS already works.

Then, inside the container, you need to create a server extension that, from time polls the user information from JH and updates the token file (I would say 1min before the token expires). The following call (inside the user container) will always return a JSON with a valid access token (unless the session expired):

```bash
curl -H "Authorization: token $JUPYTERHUB_API_TOKEN" $JPY_HUB_API_URL/users/$USER
```

PR #26, opened 2021-01-20
typo

1kastner:


1kastner:
Is it still alive?

diocas:
It is alive, but this was such a small change that I was planning to put it together with something else. But let's just merge and close. Thank you!

PR #30, opened 2021-02-19
Exchange tokens

### LINKED ISSUE #29: Authenticator: exchange tokens
To securely access EOS (and other services), we need to exchange our token with others dedicated to those services.
We need to add a configuration option to our authenticator with a list of tokens we need and we need to keep them updated just like we do for our ones.

https://auth.docs.cern.ch/user-documentation/oidc/exchange-for-api/

diocas:
This PR allows exchanging our SWAN oauth token for tokens of other services (like EOS and CERNBox).
Closes #29

### DIFF
--- a/KeyCloakAuthenticator/README.md
+++ b/KeyCloakAuthenticator/README.md
@@ -28,11 +28,6 @@ c.KeyCloakAuthenticator.username_key = 'preferred_username'
 c.KeyCloakAuthenticator.logout_redirect_uri = 'https://cern.ch/swan'
 c.KeyCloakAuthenticator.oauth_callback_url = 'https://swan.cern.ch/hub/oauth_callback'
 
-# Retrieve the user uid from the token
-def get_uid_hook(spawner, auth_state):
-    spawner.user_uid = auth_state['oauth_user']['cern_uid']
-c.KeyCloakAuthenticator.get_uid_hook = get_uid_hook
-
 # Specify the issuer url, to get all the endpoints automatically from .well-known/openid-configuration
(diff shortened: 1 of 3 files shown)


KeyCloakAuthenticator/README.md
```diff
@@ -39,6 +40,8 @@ c.KeyCloakAuthenticator.oidc_issuer = 'https://auth.cern.ch/auth/realms/cern'
 c.KeyCloakAuthenticator.accepted_roles = set()
 # Specify the role to set a user as admin
 c.KeyCloakAuthenticator.admin_role = 'swan-admin'
+# Exchange the token for tokens usable on other services (pass the audience/app id of the other services)
+c.KeyCloakAuthenticator.exchange_tokens = ['eos-service', 'cernbox-service']
```
etejedor, line 43: So you can get a token for these other services because you already have a token for SWAN, and they are all issued by the same entity?

diocas: Yes, but you can only do that if our application is approved by the owner of the other services. At CERN, this is done in the application portal (like that case where a guy asked us to exchange tokens with his apps, remember?).

etejedor: Yes ok thanks for the explanation! Do we reciprocally approve any other service in SWAN?

etejedor (approved):
The logic seems legit, from what I can tell with a limited knowledge of this token exchange API :)

KeyCloakAuthenticator/README.md
diocas: Currently we do not approve anyone.

PR #34, opened 2021-03-15
check JWT token signature and use PyJWT 2

### LINKED ISSUE #31: Validate user token in authenticator
Right now we do not validate the user oauth token. We should make this optional and check it by default.

Currently we do:
```
jwt.decode(token, verify=False, algorithms='RS256')
```
Besides not checking the token, I believe we're not compatible with the new version of PyJWT.

To read the claim without validation, we should be doing:
```
jwt.decode(token, options={"verify_signature": False}, algorithms='RS256')
```

To validate, we should be doing something like:
```
import jwt
from jwt.algorithms import RSAAlgorithm
public_key = RSAAlgorithm.from_jwk(MY_KEY)
jwt.decode(token, public_key, algorithms='RS256')
```

To get the key, we need to:
1. Extract the url `jwks_uri` from [.well-known/openid-configuration](https://auth.cern.ch/auth/realms/cern/.well-known/openid-configuration)
2. Retrieve the json from the [url in 1.](https://auth.cern.ch/auth/realms/cern/protocol/openid-connect/certs)

### LINKED ISSUE #32: KeyCloakAuthenticator compatible with pyjwt 2
Just like in https://github.com/swan-cern/jupyter-extensions/commit/edf6474b19d2c6529ff3efa9333dac2472af621c#diff-c0cd809d7ee90152667f4d7592ecd02973545b745eadf304649182a528e1b303

We also need to update the requirements to force version 2.

rccern:
fixes #31
fixes #32

### DIFF
--- a/KeyCloakAuthenticator/keycloakauthenticator/auth.py
+++ b/KeyCloakAuthenticator/keycloakauthenticator/auth.py
@@ -9,6 +9,7 @@ from oauthenticator.generic import GenericOAuthenticator
 from tornado import gen, web
 from traitlets import Unicode, Set, Bool, Any, List
 import jwt, os, pwd, time, json
+from jwt.algorithms import RSAAlgorithm
 from urllib import request, parse
 from urllib.error import HTTPError
 
@@ -61,7 +62,7 @@ class KeyCloakAuthenticator(GenericOAuthenticator):
         config=True,
         help="Users with this role login as jupyterhub administrators"
     )
-        
+
     pre_spawn_hook = Any(
         help="""
         Function to execute before spawning a session. Usefull to inject extra variables, like the oauth tokens
@@ -71,6 +72,19 @@ class KeyCloakAuthenticator(GenericOAuthenticator):
             c.KeyCloakAuthenticator.pre_spawn_hook = pre_spawn_hook
         """
     ).tag(config=True)
+
+    check_signature = Bool(
+        default_value=True,
+        config=True,
+        help="If False, it will disable JWT signature verification."
+    )
+
+    jwt_signing_algorithms = List (
+        Unicode(),
+        default_value=["HS256", "RS256"],
+        config=True,
+        help="The algorithms that can be used to check jwt signatures."
+    )
 
     exchange_tokens = List (
         Unicode(),
@@ -93,7 +107,7 @@ class KeyCloakAuthenticator(GenericOAuthenticator):
         try:
             with request.urlopen('%s/.well-known/openid-configuration' % self.oidc_issuer) as response:
                 data = json.loads(response.read())
-                
+
                 if not set(['authorization_endpoint', 'token_endpoint', 'userinfo_endpoint', 'end_session_endpoint']).issubset(data.keys()):
                     raise Exception('Unable to retrieve OIDC necessary values')
 
@@ -101,17 +115,36 @@ class KeyCloakAuthenticator(GenericOAuthenticator):
                 self.token_url = data['token_endpoint']
                 self.userdata_url = data['userinfo_endpoint']
                 self.end_session_url = data['end_session_endpoint']
-        
+
+                if self.config.check_signature :
+                    jwks_uri = data['jwks_uri']
+                    with request.urlopen(jwks_uri) as jkws_response:
+                        jwk_data = json.loads(jkws_response.read())
+                        self.public_key = RSAAlgorithm(RSAAlgorithm.SHA256).from_jwk(jwk_data['keys'][0])
+                        self.log.info('aquired public key from %s' % jwks_uri)
+
         except HTTPError:
             self.log.error("Failure to retrieve the openid configuration")
             raise
 
 
+
     def _validate_roles(self, user_roles):
         return not self.accepted_roles or (self.accepted_roles & user_roles)
 
-    def _decode_token(self, token):
-        return jwt.decode(token, verify=False, algorithms='RS256')
+    def _decode_token(self, token, options={}):
+        if not self.config.check_signature:
+            options.update({"verify_signature": False})
+        #if not explicitly disabled, verify issuer
+        options.setdefault("verify_iss", True)
+
+        try:
+            decoded_token = jwt.decode(token, self.public_key, options=options, audience=self.client_id,
+                    issuer=self.oidc_issuer, algorithms=self.jwt_signing_algorithms)
+            return decoded_token
+        except jwt.exceptions.ExpiredSignatureError:
+            self.log.info("Token expired")
+            return None
 
     def get_roles_for_token(self, token):
         decoded_token = self._decode_token(token)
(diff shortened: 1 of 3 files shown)


KeyCloakAuthenticator/keycloakauthenticator/auth.py
```diff
@@ -93,25 +100,36 @@ def __init__(self, **kwargs):
...
             raise
 
 
+
     def _validate_roles(self, user_roles):
         return not self.accepted_roles or (self.accepted_roles & user_roles)
 
-    def _decode_token(self, token):
-        return jwt.decode(token, verify=False, algorithms='RS256')
+    def _decode_token(self, token, options={}):
```
diocas, line 128: I propose we remove the call to this method from `get_roles_for_token`, and put it in `authenticate`. If decoding fails, you return a 403. If not, you take the decoded token, and pass it to `get_roles_for_token`.

KeyCloakAuthenticator/keycloakauthenticator/auth.py
```diff
@@ -192,8 +210,10 @@ async def refresh_user(self, user, handler=None):
             # Retrieve user authentication info, decode, and check if refresh is needed
             auth_state = await user.get_auth_state()
 
-            decoded_access_token = self._decode_token(auth_state['access_token'])
-            decoded_refresh_token = self._decode_token(auth_state['refresh_token'])
+            decoded_access_token =self._decode_token(auth_state['access_token'],options={"verify_exp": False})
+            # no verification of the refresh token signature as it is not needed, the auth server
+            # verifies it
+            decoded_refresh_token = self._decode_token(auth_state['refresh_token'],options={"verify_signature": False})
```
diocas, line 216: If the `refresh_token` has expiration, does the decoding work or should we also set `verify_exp False`?

diocas: (add a space in between `decoded_access_token =self._decode_token` and  `['access_token'],options`.

KeyCloakAuthenticator/keycloakauthenticator/auth.py
```diff
@@ -93,25 +100,36 @@ def __init__(self, **kwargs):
...
                 self.userdata_url = data['userinfo_endpoint']
                 self.end_session_url = data['end_session_endpoint']
-        
+
+                if self.config.check_signature :
+                    jwks_uri = data['jwks_uri']
+                    with request.urlopen(jwks_uri) as jkws_response:
+                        jwk_data = json.loads(jkws_response.read())
+                        self.public_key = RSAAlgorithm(RSAAlgorithm.SHA256).from_jwk(jwk_data['keys'][0])
+                        self.log.info('aquired public key from %s  ' % jwks_uri)
```
diocas, line 117: `'Aquired public key from %s'`

diocas (changes requested):
Just some quick changes and a question. Thanks

KeyCloakAuthenticator/keycloakauthenticator/auth.py
rccern: i checked and the refresh tokens have no expiration

KeyCloakAuthenticator/keycloakauthenticator/auth.py
rccern: if we do this, we need to change to the jhub config, because we do 
`spawner.user_roles = authenticator.get_roles_for_token(auth_state['access_token'])` now

KeyCloakAuthenticator/keycloakauthenticator/auth.py
diocas: Read my comment bellow:
` # If we request the offline_access scope, our refresh token won't have expiration`

So, it can have expiration.

KeyCloakAuthenticator/keycloakauthenticator/auth.py
diocas: This was not fixed.

KeyCloakAuthenticator/keycloakauthenticator/auth.py
rccern: but how do you know if they have the offline_access scope if you don't decode them?

diocas: Why do we care? You just need to be sure that we have `{"verify_exp": False}` just like in the previous line. Btw, this option should be False by default, correct? So in the previous line we should also not need it?

KeyCloakAuthenticator/keycloakauthenticator/auth.py
```diff
@@ -93,25 +107,44 @@ def __init__(self, **kwargs):
...
                 self.token_url = data['token_endpoint']
                 self.userdata_url = data['userinfo_endpoint']
                 self.end_session_url = data['end_session_endpoint']
-        
+
+                if self.config.check_signature :
+                    jwks_uri = data['jwks_uri']
+                    with request.urlopen(jwks_uri) as jkws_response:
+                        jwk_data = json.loads(jkws_response.read())
+                        self.public_key = RSAAlgorithm(RSAAlgorithm.SHA256).from_jwk(jwk_data['keys'][0])
```
diocas, line 123: I have a doubt here: is this generic? Since we are setting the signing algorithms, I'm not sure this will also influence.. To be seen in the future if someone complains.

KeyCloakAuthenticator/keycloakauthenticator/auth.py
```diff
@@ -191,9 +228,10 @@ async def refresh_user(self, user, handler=None):
...
-
-            decoded_access_token = self._decode_token(auth_state['access_token'])
-            decoded_refresh_token = self._decode_token(auth_state['refresh_token'])
+            # No need to verify the token, as this was verified when we received it
+            decoded_access_token = self._decode_token(auth_state['access_token'], options={"verify_signature": False})
+            # No verification of the refresh token signature is needed, the auth server already does it
+            decoded_refresh_token = self._decode_token(auth_state['refresh_token'], options={"verify_signature": False})
 
             diff_access = decoded_access_token['exp'] - time.time()
             # If we request the offline_access scope, our refresh token won't have expiration
```
diocas, line 237: I also have a doubt whether we should check the token after the refresh...

rccern: this check is not related to this change, no?  if we check every token we decode (except the refresh token, which we can't check), we are fine

diocas: But we do not decode the token after getting the refreshed one. That's my point.

rccern: ok now i'm  calling decode_token(access_token) after the refresh

diocas (approved):

Issue #35, opened 2021-04-13
Problems with get_roles_for_token


rmalik2018:
We trying to use this module for our keycloak/jupyter authentication. This module works well in a setup where roles are defined in the keycloak. However, we dynamically update user attributes with role names (List<String>). Then we use Client Mapper to pick multivalued User Attribute. Here we pass the roles to KeycloakAuthenticator.

As soon as we create new "Mapper" in Client, it overwrites all the resource_access and it removes existing system roles too. It takes out parameters which are used by other parts of jupyter such as resource_access.broker.token(its used for provider token access). I can see why it is doing it, because the newly created mapper has higher priority and it uses Token Claim Name resouce_access.{client_id}.roles. One solution could be to read the roles from an alternative Token Claim name such as roles_claim_key. 

Is it possible to provide an update to have a roles_claim_key as configurable parameter; the default can be resource_access.{client_id}.roles . However, if we should be allowed change to  for example custom.roles. This would be massively help us.

This functionality exists on 

https://github.com/jupyterhub/oauthenticator/blob/f830d4b790cd2fa6e5a2f5d6ae9b1ee844490944/oauthenticator/generic.py#L23

but its for groups not for roles.

I think the below function needs updating; where you can pass either token or custom like : roles_claim_key

https://github.com/swan-cern/jupyterhub-extensions/blob/c250978c7f2666c0caca7861d98c0747574f218f/KeyCloakAuthenticator/keycloakauthenticator/auth.py#L149

    def get_roles_for_token(self, token):
        decoded_token = self._decode_token(token)
        return set(
            decoded_token.\
               get('resource_access', {'app': ''}).\
               get(self.client_id, {'roles_list': ''}).\
               get('roles', 'no_roles')
        )

Thanks for your help.

diocas:
Hi @rmalik2018 
Currently we don't have that much time to do it, maybe in a couple of weeks... Sorry. How urgent is this for you?
You also have the possibility to open a PR with the change and we will merge it.
Cheers

rmalik2018:
Looks good me, if you would like we can test it.

PR #36, opened 2021-04-13
Problems with get_roles_for_token

rmalik2018:
Hello,

I have added one more configurable parameter for claim_roles_key as per the issue : https://github.com/swan-cern/jupyterhub-extensions/issues/35

Regards,

Rakesh Malik


diocas:
Hi @rmalik2018 sorry for the delay. This way seems a bit hardcoded and I don't this this is the best option. I would prefer something more configurable, like giving a path inside the object, and use it to fetch the value.

diocas:
Replaced by #37

PR #37, opened 2021-04-28
KeycloakAuthenticator: configurable roles retrieval

### LINKED ISSUE #35: Problems with get_roles_for_token

diocas:
Replace accepted_roles by allowed_roles and make it a list
Validate callables

Closes #35 
Replaces #36

### DIFF
--- a/KeyCloakAuthenticator/keycloakauthenticator/auth.py
+++ b/KeyCloakAuthenticator/keycloakauthenticator/auth.py
@@ -7,7 +7,7 @@ from jupyterhub.handlers import LogoutHandler
 from jupyterhub.utils import maybe_future
 from oauthenticator.generic import GenericOAuthenticator
 from tornado import gen, web
-from traitlets import Unicode, Set, Bool, Any, List
+from traitlets import Unicode, Bool, List, Any, TraitError, default, validate
 import jwt, os, pwd, time, json
 from jwt.algorithms import RSAAlgorithm
 from urllib import request, parse
@@ -50,11 +50,23 @@ class KeyCloakAuthenticator(GenericOAuthenticator):
         help="URL to invalidate the SSO cookie."
     )
 
-    accepted_roles = Set (
-        Unicode(),
-        default_value=set(),
+    # Use Any instead of Callable for compatibility with Python < 3.7
+    claim_roles_key = Any (
         config=True,
-        help="The role for which the login will be accepted. Default is all roles."
+        help="""
+            Callable that receives the spawner object and the user token json (as a dict) and returns the roles set.
+            Useful for retrieving the info from a nested object.
+            By default, retrieves the roles in resource_access.{client_id}.roles.
+        """,
+    )
+
+    # Use list instead of set to allow easy configuration in Helm/Kbernetes yaml files
+    # This gets converted to a set internally
+    allowed_roles = List (
+        Unicode(),
+        default_value=[],
+        config=True,
+        help="List with the roles for which the login will be accepted. If an empty list is given (default) all users are allowed."
     )
 
     admin_role = Unicode(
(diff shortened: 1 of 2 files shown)


diocas:
@rmalik2018 as I mentioned in a previous message, I decided to do something a bit more generic that what you proposed. Can you check this is ok for you?

KeyCloakAuthenticator/keycloakauthenticator/auth.py
```diff
@@ -93,11 +105,35 @@ def pre_spawn_hook(authenticator, spawner, auth_state):
...
+                get(env.client_id, {'roles_list': ''}).\
+                get('roles', []))
+        return get_roles
+
     def __init__(self, **kwargs):
         super().__init__(**kwargs)
 
         # Force auth state so that we can store the tokens in the user dict
         self.enable_auth_state = True
+        self._allowed_roles = set(self.allowed_roles)
```
etejedor, line 136: So we change the type of `allowed_roles` from set to list, and then we create a mirror attribute which is a set?

diocas: Yes. I do this to allow the intersection of 2 sets (when checking if a user has the necessary roles to be allowed in). Do you have a better suggestion?

etejedor: Then why changing the type of the `allowed_roles` attribute in the first place?

etejedor: Also, once `allowed_roles` is set it won't change, will it? Meaning `allowed_roles` and `_allowed_roles` do not need to be synchronized.

etejedor (approved):

KeyCloakAuthenticator/keycloakauthenticator/auth.py
diocas: I'm assuming no, hope I'm not wrong...

I'm changing the attribute to ensure we can configure it via yaml files in the k8s deployment (and that doesn't support sets).

rccern (dismissed):

diocas:
Just added a comment to the code and fixed the description of one variable. Merging.

PR #43, opened 2021-08-02
Remove end_session_endpoint from required keys

### LINKED ISSUE #41: KeyCloakAuthenticator: Lack of end_session_endpoint value in OIDC config endpoint causes startup to fail
In KeyCloakAuthenticator, when using an OIDC provider that does not have `end_session_endpoint` value in the OIDC configuration endpoint, JH fails to start, even though `enable_logout` is set to false.

didithilmy:
Fix #41

Removes `end_session_endpoint` from the required keys, since if `enable_logout` is false, the `end_session_endpoint` becomes irrelevant. This supports some OIDC provider that does not include `end_session_endpoint` in the openid-configuration endpoint.


diocas (approved):
I was wondering if we should check if logout is set to true, but I think the case of allowing logout and not doing that in the endpoint, is already covered. Hope I'm not wrong.
Approving. Sorry for taking such a long time.

PR #46, opened 2021-09-14
Swanhub041

rccern:
Readding an `await` that I had removed by mistake

Removing the #changeconfig query param because we don't have swanrc anymore

### DIFF
--- a/SwanHub/swanhub/templates/home.html
+++ b/SwanHub/swanhub/templates/home.html
@@ -124,7 +124,7 @@
 
       api.stop_server(user, {
         success: function () {
-          window.location.replace('{{base_url}}spawn?changeconfig');
+          window.location.replace('{{base_url}}spawn');
         },
         error: function () {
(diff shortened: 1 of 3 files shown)


SwanHub/swanhub/templates/home.html
```diff
@@ -108,34 +108,5 @@ <h2>Named Servers</h2>
 <script type="text/javascript">
   require(["home"]);
 </script>
-{% else %}
```
diocas, line 111: This should not be removed! Otherwise there is no way to change the user session.

diocas (changes requested):
You want (/need) to remove `changeconfig` from the **spawn** handler. You don't want to remove it from the **home** handler. Otherwise there's no way you can shutdown a session.

So this PR should touch only `sparn_handler` and `handlers_configs` and you need to change the url in the home.html (`window.location.replace('{{base_url}}spawn?changeconfig');`) and that's it.

rccern:
Thanks, I fixed the issues

diocas (approved):

PR #53, opened 2021-11-18
Fix failures to refresh token 

### LINKED ISSUE #51: KeyCloakAuthenticator: Failure to refresh tokens
The current OAuth & exchange token refresh mechanism fails due to a combination of the following:

1. An expired access token (in auth_state), fails to decode and causes a failure on `refresh_user`, and hub returns a 403 on any request. Although the refresh token is still valid and refresh can succeed.
  The following error is triggered (by returning `None` [here](https://github.com/swan-cern/jupyterhub-extensions/blob/43f02c8c6f7f2f1e9b060e040a19eebc73f62422/KeyCloakAuthenticator/keycloakauthenticator/auth.py#L183)).
  ```python
       [E 2021-11-17 10:50:06.185 SWAN auth:327] Failed to refresh the oAuth tokens
        Traceback (most recent call last):
         File "/usr/local/lib/python3.6/site-packages/keycloakauthenticator/auth.py", line 296, in refresh_user
           diff_access = decoded_access_token['exp'] - time.time()
        TypeError: 'NoneType' object is not subscriptable
  ```
2. The upstream logic in the hub is, on receiving any HTTP request, call `Authenticator.refresh_user()` if the `now() - last_refresh_time` > `Authenticator.auth_refresh_age`. (i.e if refresh_user was not already called in the last 5 minutes)
3. The KeyCloakAuthenticator skips refreshing the token and reports a successful refresh to the hub if the token does not expire in at-least `Authenticator.auth_refresh_age` (5 minutes) (logic [here](https://github.com/swan-cern/jupyterhub-extensions/blob/43f02c8c6f7f2f1e9b060e040a19eebc73f62422/KeyCloakAuthenticator/keycloakauthenticator/auth.py#L277))
4. The SwanOAuthRenew script inside the user container, sleeps for till one minute before expiry of token and pings `GET /hub/api/user`. It may or may not trigger a refresh (based on above two) and sleeps till 1 minute before new tokens expiry. if refresh was not triggered (token expiry is <60 seconds away) it sleeps atleast 60 seconds and tries again. 

(The JUPYTERHUB_API_TOKEN is separate and generated by the hub, does not expire, and is always valid. The 403 is not an issue with this token)
## Example Scenario

- `T`: Token is generated at time `T` with validity of 20 minutes (by SwanOAuthRenew or some other activity)
   - `SwanOauthRenew` script inside the user container sleeps till one minute before expiry to ping the hub again with `GET /hub/api/user`

- Between `T+14` and `T+15`: Some request to hub (culler, activity, etc), triggers hub to call refresh_user on the authenticator.
  -  The KeyCloakAuthenticator decodes the token and sees that it is valid atleast for `Authenticator.auth_refresh_age` (5 minutes) so it skips refreshing the token but returns `True`.
  - The hub marks refresh as successful and stores timestamp as last_refresh_time

- `T+19`: The SwanOAuthRenew script inside the container requests `GET /hub/api/user`, the hub checks last_refresh_time, sees that `now() - last_refresh_time < Authenticator.auth_refresh_age`, i.e the token was refreshed already in the last 5 minutes. So skips calling `refresh_user` on the authenticator.
  - `SwanOauthRenew` thinks the request succeeded, and token was refreshed, but decodes the tokens returned and sees that expiry time is <60 seconds, so sleeps atleast 60 seconds to try again.

- `T+20`: The token expires

- After `T+20`:  Calls to `GET /hub/api/user` fail with 403 as refresh_user fails. All HTTP requests to hub for this user fails including posting activity.
(... SwanOAuthRenew continues to repeat every one minute, till the container is killed due to no activity posted for 6 hours)

Depending on wether the user was idle or not, and other activity from the single user server, the above scenario only occurs after a few hours randomly.
 
## Proposed Solution
1. Refresh token even if current access token is expired without failure. Although we never arrive in this state, but this gracefully recovers if we fail to refresh in time.
2. Remove logic in the authenticator that skips refresh (no. 3) if token is valid for atleast `Authenticator.auth_refresh_age`

krishnan-r:
(The previous PR was reverted, recreating this for iterating on the fix)
- Do not skip refresh if token is current token is still valid for auth_refresh_age. This conflicts with the upstream logic to refresh only every auth_refresh_age
- Do not fail to refresh if current token is expired
- TODO: Increase Authenticator.auth_refresh_age to prevent duplicate refreshes

Fixes #51

### DIFF
--- a/KeyCloakAuthenticator/README.md
+++ b/KeyCloakAuthenticator/README.md
@@ -65,8 +65,8 @@ c.KeyCloakAuthenticator.pre_spawn_hook = pre_spawn_hook
 c.KeyCloakAuthenticator.check_signature=True
 c.KeyCloakAuthenticator.jwt_signing_algorithms = ["HS256", "RS256"]
 
-# By default jupyterhub does not trigger a refresh in `Authenticator.auth_refresh_age` seconds (default 5 minutes)
-# If you want to refresh the token less often:
+# Once a token is refreshed, by default jupyterhub does not trigger a refresh again (triggered when receiving any authenticated request) in `Authenticator.auth_refresh_age` seconds (default 5 minutes)
+# If you want to refresh the token less often, and align the refresh to your tokens expiration, this value can be changed:
 c.KeyCloakAuthenticator.auth_refresh_age = 900 # 15 minutes
 ```
 
--- a/KeyCloakAuthenticator/README.md
+++ b/KeyCloakAuthenticator/README.md
@@ -64,7 +64,12 @@ c.KeyCloakAuthenticator.pre_spawn_hook = pre_spawn_hook
 #Configure token signature verification
 c.KeyCloakAuthenticator.check_signature=True
 c.KeyCloakAuthenticator.jwt_signing_algorithms = ["HS256", "RS256"]
+
+# By default jupyterhub does not trigger a refresh in `Authenticator.auth_refresh_age` seconds (default 5 minutes)
+# If you want to refresh the token less often:
+c.KeyCloakAuthenticator.auth_refresh_age = 900 # 15 minutes
 ```
+
 
 It's also necessary to configure the Client ID and secret. One way of doing this is by setting the following environment variables:
 
--- a/KeyCloakAuthenticator/README.md
+++ b/KeyCloakAuthenticator/README.md
@@ -66,7 +66,7 @@ c.KeyCloakAuthenticator.check_signature=True
 c.KeyCloakAuthenticator.jwt_signing_algorithms = ["HS256", "RS256"]
 
 # Once a token is refreshed, by default jupyterhub does not trigger a refresh again (triggered when receiving any authenticated request) in `Authenticator.auth_refresh_age` seconds (default 5 minutes)
-# If you want to refresh the token less often, and align the refresh to your tokens expiration, this value can be changed:
+# If you want to refresh the token less often, and align the refresh to your tokens expiration, which will also trigger the update of the oAuth/OIDC token, this value can be changed:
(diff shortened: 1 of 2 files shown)


diocas:
@krishnan-r we can merge this, right? Can you add some information on the extension's README explaining how this works and how to configure JH's refresh time (needs to be close to the token lifetime)?

KeyCloakAuthenticator/README.md
```diff
@@ -64,8 +64,13 @@ c.KeyCloakAuthenticator.pre_spawn_hook = pre_spawn_hook
 #Configure token signature verification
 c.KeyCloakAuthenticator.check_signature=True
 c.KeyCloakAuthenticator.jwt_signing_algorithms = ["HS256", "RS256"]
+
+# By default jupyterhub does not trigger a refresh in `Authenticator.auth_refresh_age` seconds (default 5 minutes)
+# If you want to refresh the token less often:
```
diocas, line 69:
```suggestion
# If you want to refresh the token less often:
```
```suggestion
# Jupyterhub checks user's authentication validity every `Authenticator.auth_refresh_age` seconds (default 5 minutes), which will trigger the update of the oAuth/OIDC token as well.
# To better align this refresh rate to your token expiration, this value can be changed:
```
Something a bit clearer?

krishnan-r: I've reworded it slightly differently
I did not want to say it checks validity **every** auth_refresh_age seconds, as the check/refresh only happens when a HTTP request is received, i.e "if there was no refresh in the last auth_refresh_age seconds refresh now before handling this request"

diocas: Ya, true..

diocas (approved):

PR #61, opened 2022-03-08
KeyCloakAuthenticator v3.0.0

krishnan-r:


rccern (approved):

diocas:
@krishnan-r @rccern please don't do PR's for releases. Github is not gitlab, so when you merge - even with rebase - the original commit is not preserved. This means we have a release pointing to a commit that doesn't exist in the history.
I'll add an action to do the process from the UI.

PR #62, opened 2022-03-08
WIP Notifications

rccern:
still missing the Dockerfile

### DIFF
--- a/SwanNotificationsService/Dockerfile
+++ b/SwanNotificationsService/Dockerfile
@@ -3,6 +3,6 @@ FROM gitlab-registry.cern.ch/linuxsupport/cs9-base:20220301-1.x86_64
 RUN dnf install -y python3-pip && dnf clean all
 
 #no version here as we will likely want to build the latest
-RUN pip install swannotificationsservice
+RUN pip install --no-cache-dir swannotificationsservice
 
 ENTRYPOINT ["/usr/local/bin/swannotificationsservice" ]
--- a/SwanNotificationsService/Dockerfile
+++ b/SwanNotificationsService/Dockerfile
@@ -1,0 +1,8 @@
+FROM gitlab-registry.cern.ch/linuxsupport/cs9-base:20220301-1.x86_64
+
+RUN dnf install -y python3-pip && dnf clean all
+
+#no version here as we will likely want to build the latest
+RUN pip install swannotificationsservice
+
+ENTRYPOINT ["/usr/local/bin/swannotificationsservice" ]
--- a/SwanNotificationsService/swannotificationsservice/app.py
+++ b/SwanNotificationsService/swannotificationsservice/app.py
@@ -22,8 +22,6 @@ def main():
             (prefix, SwanNotificationsService,
                 {'notifications_file': args.notification, 'maintenance_file': args.maintenance}),
             (url_path_join(prefix,"oauth_callback"),HubOAuthCallbackHandler),
-            (r'.*', SwanNotificationsService,
-                {'notifications_file': args.notification, 'maintenance_file': args.maintenance}),
             ],
             cookie_secret=os.urandom(32))
     app.listen(args.port)
--- a/SwanNotificationsService/swannotificationsservice/app.py
+++ b/SwanNotificationsService/swannotificationsservice/app.py
@@ -1,6 +1,8 @@
 import os, argparse
 import tornado.ioloop
 import tornado.web
+from jupyterhub.utils import url_path_join
+from jupyterhub.services.auth import HubOAuthCallbackHandler
 from .service import SwanNotificationsService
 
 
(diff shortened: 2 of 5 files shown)


rccern:
added the dockerfile

SwanNotificationsService/Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM gitlab-registry.cern.ch/linuxsupport/cs9-base:20220301-1.x86_64
+
+RUN dnf install -y python3-pip && dnf clean all
+
+#no version here as we will likely want to build the latest
+RUN pip install swannotificationsservice
```
krishnan-r, line 6:
```suggestion
RUN pip install --no-cache-dir swannotificationsservice
```

SwanNotificationsService/swannotificationsservice/service.py
```diff
@@ -5,35 +5,43 @@
...
-                    'level': 'notice',
-                    'dismissible': 0,
-                    'message': 'This machine is scheduled for maintenance. Please finish your session, at your earliest convenience,'
-                               ' and visit swan.cern.ch to open a new one.<br>' + maintenance.read().replace('\n', '<br>')
-                })
+        user = self.get_current_user()
+
+        username= user['name']
+
+        user_pod = self.v1.read_namespaced_pod(f'jupyter-{username}', self.namespace)
```
krishnan-r, line 33: should we wrap this in `IOLoop.current().run_in_executor` as it blocks the thread for all other requests [as described here](https://www.tornadoweb.org/en/stable/faq.html#why-isn-t-this-example-with-time-sleep-running-in-parallel). It would not be noticable until we have many many users though

SwanNotificationsService/Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM gitlab-registry.cern.ch/linuxsupport/cs9-base:20220301-1.x86_64
```
diocas, line 1: I don't think we need anything special from centos (at least not at this time). So I would use the smallest image with Py3 available instead.

SwanNotificationsService/swannotificationsservice/app.py
```diff
@@ -16,7 +18,13 @@ def main():
 
     prefix = os.environ.get('JUPYTERHUB_SERVICE_PREFIX', args.prefix)
 
-    app = tornado.web.Application([(prefix, SwanNotificationsService, 
-                    {'notifications_file': args.notification, 'maintenance_file': args.maintenance})])
+    app = tornado.web.Application([
+            (prefix, SwanNotificationsService,
+                {'notifications_file': args.notification, 'maintenance_file': args.maintenance}),
+            (url_path_join(prefix,"oauth_callback"),HubOAuthCallbackHandler),
+            (r'.*', SwanNotificationsService,
```
diocas, line 25: Why do we need this wildcard?

SwanNotificationsService/swannotificationsservice/service.py
diocas: Careful, this is not compatible with named servers. Can we just add a query parameter to specify a named server?
This also assumes our pods are called `jupyter-xxx`, which is the case, but this is actually configurable.

SwanNotificationsService/swannotificationsservice/service.py
```diff
@@ -5,35 +5,43 @@
...
+        user_node = user_pod.spec.node_name
+
+        if self.v1.read_node(user_node).spec.unschedulable:
```
diocas, line 36: Out of curiosity, how do you define a non schedulable node?

diocas (changes requested):

SwanNotificationsService/Dockerfile
rccern: ok, please let me know which image you want to use

SwanNotificationsService/swannotificationsservice/service.py
rccern: `kubectl cordon <node>`  manually, but IIUC kubernetes itself can mark a node as unschedulable if e.g. there is no disk space, no memory available..

SwanNotificationsService/swannotificationsservice/service.py
rccern: true, I don't know how to make it compatible with named servers... in the browser we don't have this info

SwanNotificationsService/swannotificationsservice/app.py
rccern: we don't, I'll remove it

SwanNotificationsService/swannotificationsservice/service.py
diocas: We overcommit for memory, right? So, it would not be possible to add another pod there which is fine. But what happens - if there is no more memory in a node - to the running pods? Would we need to tell them that the node is going to be rebooted?

SwanNotificationsService/swannotificationsservice/service.py
diocas: No, but we need to add. Unless we send the same message to all of the user containers saying smth like "careful, your session  named X is in a node that's going to be rebooted". In this case this py code need to check for all containers of the user, not just the default.

rccern: ok i'll leave that for another pr

SwanNotificationsService/swannotificationsservice/service.py
rccern: If the node is unschedulable there is for sure an issue with the node, so I think we can always notify the user

SwanNotificationsService/Dockerfile
diocas: `python:3.9-alpine`? Sometime alpine versions are not recommended, but this is such a small/simple thing that I would expect it to work.

diocas:
@rccern how do we automate the publishing and versioning of this?

SwanNotificationsService/Dockerfile
rccern: with alpine it doesn't work, with `python:3.9` it's twice the size than with cs9, i'm leaving cs9

rccern:
we don't automate this, i don't think we will do this many times. I added a few lines in the README on how to build and push images

if you want to take care of this, please open a new PR

diocas (approved):

PR #67, opened 2022-09-19
Add Unit Tests to KeyCloakAuthenticator

krishnan-r:
Here is an attempt at adding some unit tests using `pytest`. 

This will in future help us make changes like upgrading python in the jupyterhub image with more confidence if we test against different python versions.

Suggestions/ideas welcome
TODO run these on CI


KeyCloakAuthenticator/keycloakauthenticator/tests/test_auth.py
```diff
@@ -0,0 +1,128 @@
...
+        },
+        key=private_key,
+        algorithm="RS256",
+        headers={"kid": "dummy-key-id"},
+    )
+
+
+@pytest.mark.asyncio
+async def test_refresh_user(monkeypatch):
+
```
etejedor, line 34: Thanks for writing these tests, they obviously won't hurt.

Could you add some docstrings to the two tests (and perhaps even to the helper functions above) to explain a bit what they test and how? Do you think we have a good coverage with these tests of the cases our authenticator handles?

krishnan-r: I've added the docstrings and some comments.

Regarding the coverage, this only covers the `refresh_user` method at the moment which was the one I was debugging issues with. I will add more test cases in future as separate PRs

krishnan-r:
@etejedor Sorry for the delay, I've updated the PR, and also added a CI workflow for running the tests on PRs

.github/workflows/check-KeyCloakAuthenticator.yml
```diff
@@ -0,0 +1,30 @@
...
+
+jobs:
+  check_unit_tests:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+
+      - uses: actions/setup-python@v4
+        with:
+          python-version: '3.10'
```
etejedor, line 17: Can this be defined as a variable somewhere, so that we could potentially reuse it in other unit tests for other components / extensions? It should be a variable that contains the python version we currently use in production for jupyterhub (and jupyter? ideally the same).

krishnan-r: Good point. That [version](https://github.com/swan-cern/jupyterhub-image/blob/69a56c2c8c89926d17c401daf638c5ce005aed6c/Dockerfile#L60) is 3.6 currently for the hub image and 3.9 I think for user image.

I think we should also test against the latest version that we want to update to.

I will check to setup `tox` or a similar tool that runs the tests on a matrix of versions perhaps, including the current production version?

etejedor: I would align the hub and the user jupyter versions, and test all our components with that one. What's the advantage of testing multiple Python versions if in production we will just run one? But we need a way to set the Python version for all or our extensions/components.

.github/workflows/check-KeyCloakAuthenticator.yml
```diff
@@ -14,7 +14,7 @@ jobs:
 
       - uses: actions/setup-python@v4
         with:
-          python-version: '3.10'
+          python-version-file: '.python-version'
```
etejedor, line 17: Nice, thanks!

etejedor (approved):

krishnan-r:
@etejedor I added a commit that fixes the failing dependency resolution for Python 3.6.8, can you review again, thanks

etejedor (approved):

PR #68, opened 2022-11-04
Add Prometheus Metrics and Duration Logs to Authenticator

krishnan-r:


KeyCloakAuthenticator/keycloakauthenticator/auth.py
```diff
@@ -200,29 +202,32 @@ async def _exchange_tokens(self, token):
...
+
+                req = HTTPRequest(
+                    self.token_url,
+                    method="POST",
+                    headers=self._get_headers(),
+                    body=data,
+                )
+                response = await self.fetch(req, "exchanging token")
+                tokens[new_token] = response.get('access_token', None)
+                self.log.info('Exchanged {} token in {} seconds'.format(new_token, time.time() - start))
```
etejedor, line 225: I was wondering whether we need the time in the logs at all, or we can just live with the histograms for diagnosis.

KeyCloakAuthenticator/keycloakauthenticator/auth.py
```diff
@@ -200,29 +202,32 @@ async def _exchange_tokens(self, token):
...
-            req = HTTPRequest(
-                self.token_url,
-                method="POST",
-                headers=self._get_headers(),
-                body=data,
-            )
-            response = await self.fetch(req, "exchanging token")
-            tokens[new_token] = response.get('access_token', None)
-
+            with metric_exchange_token.time():
```
etejedor, line 205: So here all the exchanges will go into the same histogram. It would be nice to have one histogram per service we exchange tokens with. Can these histograms be parameterized in some way? Otherwise, for the exchanges, we can have a dictionary `metric_exchange_token` that contains `{ 'eos' : histo_eos, 'cernbox' : histo_cernbox }` and so on. This means that the constructor of `KeyCloakAuthenticator` needs to initialize that dictionary, since at that point you know the `exchange_tokens`.

KeyCloakAuthenticator/keycloakauthenticator/metrics.py
```diff
@@ -0,0 +1,37 @@
+from prometheus_client import Histogram
```
etejedor, line 1: It would be nice to have a module docstring here before any code, to explain what's the purpose of this module and what variables it defines.

KeyCloakAuthenticator/keycloakauthenticator/metrics.py
```diff
@@ -0,0 +1,37 @@
...
+    15.0,
+    20.0,
+    30.0,
+    float("inf"),
+)
+
+
+_DURATION_SECONDS = Histogram(
+    "keycloak_authenticator_duration_seconds",
+    "Histogram of KeyCloakAuthenticator durations",
```
etejedor, line 28:
```suggestion
    "Histogram of KeyCloakAuthenticator method times",
```

KeyCloakAuthenticator/keycloakauthenticator/metrics.py
```diff
@@ -0,0 +1,37 @@
...
+    10.0,
+    15.0,
+    20.0,
+    30.0,
+    float("inf"),
+)
+
+
+_DURATION_SECONDS = Histogram(
+    "keycloak_authenticator_duration_seconds",
```
etejedor, line 27:
```suggestion
    "keycloak_authenticator_method_times",
```

KeyCloakAuthenticator/keycloakauthenticator/auth.py
krishnan-r: I think it is still worth logging individual times, as it might help us debug issues for individual users.

This does not increase the volume of our logs significantly, most of the cases here I've only appended the time to an existing log message

KeyCloakAuthenticator/keycloakauthenticator/metrics.py
krishnan-r: done

KeyCloakAuthenticator/keycloakauthenticator/auth.py
krishnan-r: I've split this into two metrics now, 'method' and 'request' duration, with a label describing each request.

KeyCloakAuthenticator/keycloakauthenticator/metrics.py
krishnan-r: For this and the suggestion above, I'm inclined to keep this as `duration_seconds` following the convention of other metrics in prometheus (https://prometheus.io/docs/practices/naming/)

The `_seconds` describes the unit of the metric

KeyCloakAuthenticator/keycloakauthenticator/auth.py
etejedor: It might not increase a lot the number of logs, but it makes us introduce time measures mixed with our code.

In what circumstances would you want to see the times for an individual user? Is it feasible that we get different times for refreshing or exchanging tokens depending on the user, or it will just depend on the availability / performance of the servers that provide those tokens?

KeyCloakAuthenticator/keycloakauthenticator/auth.py
```diff
@@ -200,29 +202,32 @@ async def _exchange_tokens(self, token):
...
-            req = HTTPRequest(
-                self.token_url,
-                method="POST",
-                headers=self._get_headers(),
-                body=data,
-            )
-            response = await self.fetch(req, "exchanging token")
-            tokens[new_token] = response.get('access_token', None)
-
+            with metric_exchange_token.labels("exchange_token_{}".format(new_token.replace("-","_"))).time():
```
etejedor, line 205: Can you add a comment before this line that explains what you are doing here? You are emitting a different time for each of the services we exchange token with, and you achieve that by setting a value for the label `request` of the histogram?

KeyCloakAuthenticator/keycloakauthenticator/auth.py
krishnan-r: I was thinking this info might help us drill down, with the metrics we only see an aggregated picture, not if a particular user was facing slowness.

Do you suggest just keeping the prometheus metrics then and removing the log timings?

etejedor: Ok, let's leave the timings for now. But I'd like to get rid of them in the future if we realize they are not really necessary.

etejedor (dismissed):

etejedor (approved):

PR #71, opened 2022-11-15
Add a loading screen for login page.

krishnan-r:
This page shows a spinner and redirects from javascript to start the oauth process This prevents an unresponsive page when logins are slow.

- Requires Authenticator.auto_login to be disabled (https://github.com/swan-cern/swan-charts/pull/104)
- The only thing different here from puppet infrastructure is that the redirect URL does not include the hostname, as the hostname is not available (from the puppet templating)


etejedor (approved):

diocas (approved):
The hostname could be provided via config. It's not needed, but then we could re-use this in puppet as well.
But don't waste time on this :)

PR #77, opened 2022-11-30
Automate SWANTA

krishnan-r:


diocas (approved):
We need the same for the systemuser image, IIRC

PR #80, opened 2023-02-13
Log access token in case of exchange failure for debugging purposes

etejedor:
In the SWAN logs, we have observed that sporadically, for certain users, we are able to successfully complete a refresh request to the SSO for the SWAN access token, but when we use that token in the subsequent exchange requests, those fail.

This PR intends to help with the debugging of such issue by logging the user access token in case of exchange error to be able to provide it to the SSO team.


etejedor:
Test failure is unrelated to this PR. The runners got probably updated and we can't use Python 3.6 anymore (`Error: The version '3.6.8' with architecture 'x64' was not found for Ubuntu 22.04.`). To be fixed in another PR.

PR #82, opened 2023-06-16
Jh 4 0

diocas:
Missing updates of Kubespawner, that @etejedor will work on.

Ofc this still has things that need to be removed and cleaned, but we need 1st to decomission puppet (to remove the extra user api handler) and to a new image, cleaned of extra, unwanted configs.
Also, once we update the design and user a more upstream look and feel, the overload of the spawner handler is no longer needed.


etejedor:
This is to be merged, when agreed, together with the corresponding PR of the swan charts: https://github.com/swan-cern/swan-charts/pull/167.

We might need to reconsider if this needs to go together with an update of the systemuser image. Therefore, this is on hold for now.

SwanSpawner/swanspawner/swankubespawner.py
```diff
@@ -48,6 +51,40 @@ async def start(self):
             self.log.error("Error while spawning the user container: %s", e, exc_info=True)
             raise e
 
+    async def stop(self, now=False):
+        '''Do custom cleanup after terminating user pod'''
+        try:
+            await super().stop()
+        finally:
+            # Delete Kubernetes secret storing EOS kerberos ticket of the user
```
diocas, line 59: If we have local_home, do we even store this secret?

I know this will try and continue if it's not there. But it seems a bit hardcoded. Same for spark bellow. Can't we keep track of any added secret and do a for secret in created_secrets delete secret?

diocas (commented):
Approved (cannot approve my own PR)

etejedor (approved):

PR #86, opened 2023-11-27
Allow to configure an extra URL for a CentOS7-based user image

etejedor:
Since now two images will be supported (Alma9 and CentOS7), the URLs of both need to be configured.

The Alma9 URL is configured via chart settings, and the old CentOS7 image URL is configured via an option of our kubespawner. We make the spawner switch image if centos7 is found in the selected platform.

Goes together with https://github.com/swan-cern/swan-charts/pull/192

### DIFF
--- a/SwanSpawner/swanspawner/swankubespawner.py
+++ b/SwanSpawner/swanspawner/swankubespawner.py
@@ -22,7 +22,8 @@ class SwanKubeSpawner(define_SwanSpawner_from(KubeSpawner)):
     )
 
     async def start(self):
-        """Perform the operations necessary for GPU support
+        """Perform extra configurations required for SWAN session spawning in
+        kubernetes.
         """
 
         if self._gpu_requested():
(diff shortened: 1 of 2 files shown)


SwanSpawner/swanspawner/swankubespawner.py
```diff
@@ -42,6 +48,15 @@ async def start(self):
         self.mem_limit = self.user_options[self.user_memory]
         self.mem_guarantee = ceil(self.mem_limit * self.mem_request_fraction)
 
+        # An Alma9-based user image is configured by default via the chart
+        # settings, but users could still select a CentOS7 platform.
+        # In that case, reconfigure to use a CentOS7-based user image
+        if 'centos7' in self.user_options[self.platform_field]:
+            image = self.centos7_image
+            if image:
```
diocas, line 56: not?

etejedor: Absolutely, I initially had `if image is not None` but then I saw the default of a Unicode is "", and changed it (incorrectly).

diocas (approved):

PR #88, opened 2024-01-10
Changes to the web form template related to Alma9

etejedor:
This PR adds:
- A message to the form to advertise the Alma9 platform and JupyterLab.
- The automatic disabling of the Spark cluster if the selected platform is Alma9 (see commit message for details on why).

Screenshot from 2024-01-10 22-47-25

### DIFF
--- a/SwanSpawner/swanspawner/templates/options_form_template.html
+++ b/SwanSpawner/swanspawner/templates/options_form_template.html
@@ -138,6 +138,8 @@
         // adjust clusters option for lcg
         var selectCluster = document.getElementById('clusterOptions');
         selectCluster.innerHTML = '';
+        // Make sure Spark cluster options are enabled on LCG release change
+        selectCluster.removeAttribute('disabled');
 
         for( var i = 0 ; i < lcgData.clusters.length ; i++ ){
             var lcgCluster = lcgData.clusters[i];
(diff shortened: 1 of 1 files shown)


PMax5 (dismissed):

SwanSpawner/swanspawner/templates/options_form_template.html
```diff
@@ -169,13 +170,44 @@
         }
     }
 
+    /**
+     * Modifies the selectable Spark clusters depending on the chosen platform
+     */
+    function adjust_spark() {
+        var selectPlatform = document.getElementById('platformOptions');
```
diocas, line 177:
```js
var platformOptions = document.getElementById('platformOptions');
var clusterOptions  = document.getElementById('clusterOptions');

var isAlma = platformOptions.selectedOptions[0].text.startsWith('AlmaLinux 9');
clusterOptions.disabled = !isAlma;

if (isAlma) {
    clusterOptions.selectedIndex = 0;
}
```

Our selects only allow single selections, so I would say it's safe to pick the option `[0]`, as the array will always have 1 element.
And I would simplify the rest: if you pick Alma, the Spark Cluster selection changes to None and gets disabled.

diocas (changes requested):

SwanSpawner/swanspawner/templates/options_form_template.html
```diff
@@ -137,6 +138,8 @@
         // adjust clusters option for lcg
         var selectCluster = document.getElementById('clusterOptions');
         selectCluster.innerHTML = '';
+        // Make sure Spark cluster options are enabled on LCG release change
+        selectCluster.removeAttribute('disabled');
```
etejedor, line 142: Please check the new version @diocas . I also had to add this line to make sure the Spark selectable is enabled when switching lcg stack.

diocas (approved):

PR #94, opened 2024-05-22
Custom envs

RodrigoSobral2000:
Result:
Spawner can now launch custom environments through its web interface.

Developments:
Thanks to changes on SwanSpawner and SwanHub, its now possible to avoid having to install all LCG packages and provide a specific amount of packages through cernbox/eos path or git repo containing a requirements.txt file.

This new custom virtual environment can then be launched by a generic python venv or a acc-py stack.

The user gets redirected to /customenv where the process details are shown up in real time, so the users is able to keep up with its going on under the hood. In the end, a completely empty custom environment (just with the provided packages) gets ready for the user to work on.

### DIFF
--- a/SwanHub/swanhub/handlers_configs.py
+++ b/SwanHub/swanhub/handlers_configs.py
@@ -31,16 +31,16 @@ class SpawnHandlersConfigs(SingletonConfigurable):
 
     user_script_env_field = 'scriptenv'
     
-    cernbox_pattern = Unicode(
-        default_value=r'^\/?([\w-]+\/)*[\w-]+\/?$',
+    eos_pattern = Unicode(
+        default_value=r'^(?:\$CERNBOX_HOME\/)?\/?([\w-]+\/)*[\w-]+\/?$',
         config=True,
-        help='Regular expression pattern for requirements provided by a CERNBox file.'
+        help='Regular expression pattern for requirements provided by a EOS folder.'
     )
 
-    cernbox_special_type = Unicode(
-        default_value='cernbox',
+    eos_special_type = Unicode(
+        default_value='eos',
         config=True,
-        help='Special type for requirements provided by a CERNBox folder.'
+        help='Special type for requirements provided by a EOS folder.'
     )
 
     git_pattern = Unicode(
--- a/SwanHub/swanhub/handlers_configs.py
+++ b/SwanHub/swanhub/handlers_configs.py
@@ -17,7 +17,7 @@ class SpawnHandlersConfigs(SingletonConfigurable):
     
     source_type = 'source_type'
 
-    customenv_type = 'customenv_type'
+    builder = 'builder'
     
     customenv_type_version = 'customenv_type_version'
 
--- a/SwanHub/swanhub/handlers_configs.py
+++ b/SwanHub/swanhub/handlers_configs.py
@@ -17,9 +17,9 @@ class SpawnHandlersConfigs(SingletonConfigurable):
 
     source_type = 'source_type'
 
-    customenv_type = 'customenv_type'
+    builder = 'builder'
     
-    customenv_type_version = 'customenv_type_version'
+    builder_version = 'builder_version'
             
     repository_type = 'repository_type'
 
--- a/SwanHub/swanhub/handlers_configs.py
+++ b/SwanHub/swanhub/handlers_configs.py
@@ -14,6 +14,14 @@ class SpawnHandlersConfigs(SingletonConfigurable):
     """
         Singleton class where all the configurations are stored
     """
+    
+    source_type = 'source_type'
+
+    customenv_type = 'customenv_type'
+    
+    customenv_type_version = 'customenv_type_version'
+            
+    requirements = 'requirements'
 
     software_source = 'software_source'
 
--- a/SwanHub/swanhub/handlers_configs.py
+++ b/SwanHub/swanhub/handlers_configs.py
@@ -32,7 +32,7 @@ class SpawnHandlersConfigs(SingletonConfigurable):
     user_script_env_field = 'scriptenv'
     
     eos_pattern = Unicode(
-        default_value=r'^(?:\$CERNBOX_HOME\/)?\/?([\w-]+\/)*[\w-]+\/?$',
+        default_value=r'^(?:\$CERNBOX|(?:/[^/\n]+)*/[^/\n]+)$',
         config=True,
         help='Regular expression pattern for requirements provided by a EOS folder.'
     )
--- a/SwanHub/swanhub/handlers_configs.py
+++ b/SwanHub/swanhub/handlers_configs.py
@@ -14,8 +14,8 @@ class SpawnHandlersConfigs(SingletonConfigurable):
     """
         Singleton class where all the configurations are stored
     """
-    
-    source_type = 'source_type'
+
+    software_source = 'software_source'
 
     builder = 'builder'
     
--- a/SwanHub/swanhub/handlers_configs.py
+++ b/SwanHub/swanhub/handlers_configs.py
@@ -93,12 +93,6 @@ class SpawnHandlersConfigs(SingletonConfigurable):
         help='Default platform configuration for LCG views.'
     )
 
-    notebook = Unicode(
-        default_value='notebook',
-        config=True,
-        help='Name of the notebook file to launch with its arguments'
-    )
-
     local_home = Bool(
         default_value=False,
         config=True,
--- a/SwanHub/swanhub/handlers_configs.py
+++ b/SwanHub/swanhub/handlers_configs.py
@@ -25,72 +25,24 @@ class SpawnHandlersConfigs(SingletonConfigurable):
 
     repository = 'repository'
 
+    project_folder = 'project_folder'
+
     lcg_rel_field = 'LCG-rel'
 
     platform_field = 'platform'
 
     user_script_env_field = 'scriptenv'
 
-    user_n_cores = 'ncores'
-
-    user_memory = 'memory'
-
     spark_cluster_field = 'spark-cluster'
 
-    condor_pool = 'condor-pool'
+    customenv_special_type = 'customenv'
 
-    eos_pattern = Unicode(
-        default_value=r'^(?:\$CERNBOX|(?:/[^/\n]+)*/[^/\n]+)?$',
-        config=True,
-        help='Regular expression pattern for the repository provided by a EOS folder.'
-    )
-
-    eos_special_type = Unicode(
-        default_value='eos',
-        config=True,
-        help='Special type for the repository provided by a EOS folder.'
-    )
-
-    git_pattern = Unicode(
-        default_value=r'https?://(?:github\.com|gitlab\.cern\.ch)/([^/\s]+)/([^/\s]+)/?',
-        config=True,
-        help='Regular expression pattern for the repository provided by a GitLab or GitHub repository.'
-    )
-
-    git_special_type = Unicode(
-        default_value='git',
-        config=True,
-        help='Special repository type for Git repositories.'
-    )
-
-    lcg_special_type = Unicode(
-        default_value='lcg',
-        config=True,
-        help='Special type for default sourcing.'
-    )
-
-    customenv_special_type = Unicode(
-        default_value='customenv',
-        config=True,
-        help='Special type for sourcing the environment.'
-    )
-
-    accpy_special_type = Unicode(
-        default_value='accpy',
-        config=True,
-        help='Special type for custom environments.'
-    )
+    accpy_special_type = 'accpy'
 
     env_name = Unicode(
         default_value='{project_folder}_env',
         config=True,
         help='Name format for custom environment launched by the user.'
-    )
-
-    default_platform = Unicode(
-        default_value='x86_64-el9-gcc13-opt',
-        config=True,
-        help='Default platform configuration for LCG views.'
     )
 
     local_home = Bool(
--- a/SwanHub/swanhub/handlers_configs.py
+++ b/SwanHub/swanhub/handlers_configs.py
@@ -21,7 +21,7 @@ class SpawnHandlersConfigs(SingletonConfigurable):
     
     builder_version = 'builder_version'
             
-    repository_type = 'repository_type'
+    repo_type = 'repo_type'
 
     repository = 'repository'
 
--- a/SwanHub/swanhub/handlers_configs.py
+++ b/SwanHub/swanhub/handlers_configs.py
@@ -20,8 +20,6 @@ class SpawnHandlersConfigs(SingletonConfigurable):
     customenv_type = 'customenv_type'
     
     customenv_type_version = 'customenv_type_version'
-            
-    requirements = 'requirements'
 
     software_source = 'software_source'
 
--- a/SwanHub/swanhub/handlers_configs.py
+++ b/SwanHub/swanhub/handlers_configs.py
@@ -14,45 +14,59 @@ class SpawnHandlersConfigs(SingletonConfigurable):
     """
         Singleton class where all the configurations are stored
     """
-    
+
     source_type = 'source_type'
 
     customenv_type = 'customenv_type'
     
     customenv_type_version = 'customenv_type_version'
+            
+    repository_type = 'repository_type'
 
-    requirements = 'requirements'
-    
-    requirements_type = 'requirements_type'
-    
+    repository = 'repository'
+
     lcg_rel_field = 'LCG-rel'
+
+    platform_field = 'platform'
+
+    user_script_env_field = 'scriptenv'
+
+    user_n_cores = 'ncores'
+
+    user_memory = 'memory'
 
     spark_cluster_field = 'spark-cluster'
 
-    user_script_env_field = 'scriptenv'
-    
+    condor_pool = 'condor-pool'
+
     eos_pattern = Unicode(
-        default_value=r'^(?:\$CERNBOX|(?:/[^/\n]+)*/[^/\n]+)$',
+        default_value=r'^(?:\$CERNBOX|(?:/[^/\n]+)*/[^/\n]+)?$',
         config=True,
-        help='Regular expression pattern for requirements provided by a EOS folder.'
+        help='Regular expression pattern for the repository provided by a EOS folder.'
     )
 
     eos_special_type = Unicode(
         default_value='eos',
         config=True,
-        help='Special type for requirements provided by a EOS folder.'
+        help='Special type for the repository provided by a EOS folder.'
     )
 
     git_pattern = Unicode(
         default_value=r'https?://(?:github\.com|gitlab\.cern\.ch)/([^/\s]+)/([^/\s]+)/?',
         config=True,
-        help='Regular expression pattern for requirements provided by a GitLab or GitHub repository.'
+        help='Regular expression pattern for the repository provided by a GitLab or GitHub repository.'
     )
 
     git_special_type = Unicode(
         default_value='git',
         config=True,
-        help='Special requirements type for Git repositories.'
+        help='Special repository type for Git repositories.'
+    )
+
+    lcg_special_type = Unicode(
+        default_value='lcg',
+        config=True,
+        help='Special type for default sourcing.'
     )
 
     customenv_special_type = Unicode(
@@ -71,6 +85,18 @@ class SpawnHandlersConfigs(SingletonConfigurable):
         default_value='{project_folder}_env',
         config=True,
         help='Name format for custom environment launched by the user.'
+    )
+
+    default_platform = Unicode(
+        default_value='x86_64-el9-gcc13-opt',
+        config=True,
+        help='Default platform configuration for LCG views.'
+    )
+
+    notebook = Unicode(
+        default_value='notebook',
+        config=True,
+        help='Name of the notebook file to launch with its arguments'
     )
 
     local_home = Bool(
--- a/SwanHub/swanhub/handlers_configs.py
+++ b/SwanHub/swanhub/handlers_configs.py
@@ -19,19 +19,11 @@ class SpawnHandlersConfigs(SingletonConfigurable):
 
     builder = 'builder'
     
-    customenv_type_version = 'customenv_type_version'
-
-    software_source = 'software_source'
-
-    builder = 'builder'
-    
     builder_version = 'builder_version'
             
     repository_type = 'repository_type'
 
     repository = 'repository'
-
-    project_folder = 'project_folder'
 
     lcg_rel_field = 'LCG-rel'
 
--- a/SwanHub/swanhub/handlers_configs.py
+++ b/SwanHub/swanhub/handlers_configs.py
@@ -20,14 +20,58 @@ class SpawnHandlersConfigs(SingletonConfigurable):
     customenv_type = 'customenv_type'
     
     customenv_type_version = 'customenv_type_version'
-            
-    requirements = 'requirements'
 
+    requirements = 'requirements'
+    
+    requirements_type = 'requirements_type'
+    
     lcg_rel_field = 'LCG-rel'
 
     spark_cluster_field = 'spark-cluster'
 
     user_script_env_field = 'scriptenv'
+    
+    cernbox_pattern = Unicode(
+        default_value=r'^\/?([\w-]+\/)*[\w-]+\/?$',
+        config=True,
+        help='Regular expression pattern for requirements provided by a CERNBox file.'
+    )
+
+    cernbox_special_type = Unicode(
+        default_value='cernbox',
+        config=True,
+        help='Special type for requirements provided by a CERNBox folder.'
+    )
+
+    git_pattern = Unicode(
+        default_value=r'https?://(?:github\.com|gitlab\.cern\.ch)/([^/\s]+)/([^/\s]+)/?',
+        config=True,
+        help='Regular expression pattern for requirements provided by a GitLab or GitHub repository.'
+    )
+
+    git_special_type = Unicode(
+        default_value='git',
+        config=True,
+        help='Special requirements type for Git repositories.'
+    )
+
+    customenv_special_type = Unicode(
+        default_value='customenv',
+        config=True,
+        help='Special type for sourcing the environment.'
+    )
+
+    accpy_special_type = Unicode(
+        default_value='accpy',
+        config=True,
+        help='Special type for custom environments.'
+    )
+
+    env_name = Unicode(
+        default_value='{project_folder}_env',
+        config=True,
+        help='Name format for custom environment launched by the user.'
+    )
 
     local_home = Bool(
         default_value=False,
--- a/SwanHub/swanhub/handlers_configs.py
+++ b/SwanHub/swanhub/handlers_configs.py
@@ -25,8 +25,6 @@ class SpawnHandlersConfigs(SingletonConfigurable):
 
     repository = 'repository'
 
-    project_folder = 'project_folder'
-
     lcg_rel_field = 'LCG-rel'
 
     platform_field = 'platform'
@@ -38,12 +36,6 @@ class SpawnHandlersConfigs(SingletonConfigurable):
     customenv_special_type = 'customenv'
 
     accpy_special_type = 'accpy'
-
-    env_name = Unicode(
-        default_value='{project_folder}_env',
-        config=True,
-        help='Name format for custom environment launched by the user.'
-    )
 
     local_home = Bool(
         default_value=False,
--- a/SwanHub/swanhub/handlers_configs.py
+++ b/SwanHub/swanhub/handlers_configs.py
@@ -14,6 +14,14 @@ class SpawnHandlersConfigs(SingletonConfigurable):
     """
         Singleton class where all the configurations are stored
     """
+    
+    source_type = 'source_type'
+
+    customenv_type = 'customenv_type'
+    
+    customenv_type_version = 'customenv_type_version'
+            
+    requirements = 'requirements'
 
     lcg_rel_field = 'LCG-rel'
 
--- a/SwanHub/swanhub/handlers_configs.py
+++ b/SwanHub/swanhub/handlers_configs.py
@@ -15,7 +15,7 @@ class SpawnHandlersConfigs(SingletonConfigurable):
         Singleton class where all the configurations are stored
     """
 
-    source_type = 'source_type'
+    software_source = 'software_source'
 
     builder = 'builder'
     
--- a/SwanHub/swanhub/handlers_configs.py
+++ b/SwanHub/swanhub/handlers_configs.py
@@ -31,6 +31,8 @@ class SpawnHandlersConfigs(SingletonConfigurable):
 
     repository = 'repository'
 
+    project_folder = 'project_folder'
+
     lcg_rel_field = 'LCG-rel'
 
     platform_field = 'platform'
--- a/SwanHub/swanhub/spawn_handler.py
+++ b/SwanHub/swanhub/spawn_handler.py
@@ -105,6 +105,9 @@ class SpawnHandler(JHSpawnHandler):
         for key, byte_list in self.request.files.items():
             form_options["%s_file" % key] = byte_list
 
+        if form_options[configs.source_type][0] == "customenv" and form_options[configs.requirements][0] == '':
+            raise web.HTTPError(400, "Requirements not provided")
+
         start_time_spawn = time.time()
 
         options = {}
@@ -172,7 +175,6 @@ class SpawnHandler(JHSpawnHandler):
                 user,
                 default=url_path_join(self.hub.base_url, "spawn-pending", user.escaped_name, server_name),
             )
-
         self.redirect(next_url)
 
     async def _render_form_wrapper(self, for_user, message=''):
--- a/SwanHub/swanhub/spawn_handler.py
+++ b/SwanHub/swanhub/spawn_handler.py
@@ -109,19 +109,12 @@ class SpawnHandler(JHSpawnHandler):
 
         start_time_spawn = time.time()
 
+        options = {}
         try:
             options = await maybe_future(spawner.run_options_from_form(form_options))
 
             if options.get(configs.source_type) == configs.customenv_special_type and not options.get(configs.repository):
-                raise ValueError("Requirements not provided")
-
-            # Dont allow the session to spawn if the repository are not valid
-            eos_match = re.match(configs.eos_pattern, options.get(configs.repository))
-            git_match = re.match(configs.git_pattern, options.get(configs.repository))
-            if options.get(configs.repository_type) == configs.eos_special_type and not eos_match:
-                raise ValueError(f"Invalid EOS path for requirements: {options.get(configs.repository)}")
-            if options.get(configs.repository_type) == configs.git_special_type and not git_match:
-                raise ValueError(f"Invalid Git repository for requirements: {options.get(configs.repository)}")
+                raise ValueError("Repository not provided")
 
             await self.spawn_single_user(user, server_name=server_name, options=options)
 
@@ -170,28 +163,26 @@ class SpawnHandler(JHSpawnHandler):
         if current_user is user:
             self.set_login_cookie(user)
 
-        if options.get(configs.source_type) == configs.customenv_special_type:
-            project_folder =  options.get(configs.repository).split('/')[-1]
-            if options.get(configs.repository).startswith('http'):
-                project_folder = git_match[2] if git_match else user.escaped_name
-            
+        if options.get(configs.source_type) == configs.customenv_special_type:            
             # Add the query parameters to the URL
             query_params = {
-                "env": configs.env_name.format(project_folder=project_folder),
+                "env": configs.env_name.format(project_folder=options.get(configs.project_folder)), # {reponame}_env or {lastfolder}_env
                 "repo": options.get(configs.repository),
             }
             if options.get(configs.customenv_type) == configs.accpy_special_type:
                 query_params[options.get(configs.customenv_type)] = options.get(configs.customenv_type_version)
 
+            # Execution SwanCustomEnvs extension with the corresponding query arguments
             next_url = self.get_next_url(
                 user,
                 default=url_concat(url_path_join("user", user.escaped_name, "customenvs", server_name), query_params),
             )
-        else:
+        else: # LCG release
             next_url = self.get_next_url(
                 user,
                 default=url_path_join(self.hub.base_url, "spawn-pending", user.escaped_name, server_name),
             )
+
         self.redirect(next_url)
 
     async def _render_form_wrapper(self, for_user, message=''):
@@ -246,7 +237,7 @@ class SpawnHandler(JHSpawnHandler):
                 metrics.append((metric, (date, 1)))
 
         spawn_context_key = ".".join(
-            [options.get(configs.lcg_rel_field), options.get(configs.spark_cluster_field)])
+            [options.get(configs.lcg_rel_field, "CustomEnv"), options.get(configs.spark_cluster_field, "NoSpark")])
         if not spawn_exception:
             # Add spawn success (no exception) and duration to the log and send as metrics
             spawn_exc_class = "None"
--- a/SwanHub/swanhub/spawn_handler.py
+++ b/SwanHub/swanhub/spawn_handler.py
@@ -5,6 +5,7 @@
 
 import time
 import os
+import re
 from jupyterhub.handlers.pages import SpawnHandler as JHSpawnHandler
 from jupyterhub.utils import url_path_join, maybe_future
 from jupyterhub.scopes import needs_scope
@@ -105,14 +106,23 @@ class SpawnHandler(JHSpawnHandler):
         for key, byte_list in self.request.files.items():
             form_options["%s_file" % key] = byte_list
 
-        if form_options[configs.source_type][0] == "customenv" and form_options[configs.requirements][0] == '':
-            raise web.HTTPError(400, "Requirements not provided")
-
         start_time_spawn = time.time()
 
         options = {}
         try:
             options = await maybe_future(spawner.run_options_from_form(form_options))
+
+            if options[configs.source_type] == configs.customenv_special_type and not options[configs.requirements]:
+                raise ValueError("Requirements not provided")
+
+            # Dont allow the session to spawn if the requirements are not valid
+            cernbox_match = re.match(configs.cernbox_pattern, options[configs.requirements])
+            git_match = re.match(configs.git_pattern, options[configs.requirements])
+            if options[configs.requirements_type] == configs.cernbox_special_type and not cernbox_match:
+                raise ValueError(f"Invalid CERNBox folder for requirements: {options[configs.requirements]}")
+            if options[configs.requirements_type] == configs.git_special_type and not git_match:
+                raise ValueError(f"Invalid Git repository for requirements: {options[configs.requirements]}")
+
             await self.spawn_single_user(user, server_name=server_name, options=options)
 
             # if spawn future is already done it is success,
--- a/SwanHub/swanhub/spawn_handler.py
+++ b/SwanHub/swanhub/spawn_handler.py
@@ -56,31 +56,12 @@ class SpawnHandler(JHSpawnHandler):
             self.finish(form)
             return
 
-        if not self.request.query_arguments:
-            try:
-                await super().get(for_user, server_name)
-            except web.HTTPError as e:
-                form = await self._render_form_wrapper(user, message=e.message)
-                self.finish(form)
-                return
-        else:
-            query_options = {key: [value[0].encode("utf-8")] for key, value in parse_qs(urlparse(unquote(self.request.uri)).query).items()}
-            self.request.body_arguments = {
-                configs.source_type: query_options.get(configs.source_type, [bytes(configs.lcg_special_type.encode("utf-8"))]),
-                configs.repository: query_options.get(configs.repository, [b'']),
-                configs.repository_type: query_options.get(configs.repository_type, [b'']),
-                configs.customenv_type: query_options.get(configs.customenv_type, [b'']),
-                configs.customenv_type_version: query_options.get(configs.customenv_type_version, [b'']),
-                configs.lcg_rel_field: query_options.get(configs.lcg_rel_field, [b'']),
-                configs.platform_field: query_options.get(configs.platform_field, [bytes(configs.default_platform.encode("utf-8"))]),
-                configs.spark_cluster_field: query_options.get(configs.spark_cluster_field, [b'']),
-                configs.user_script_env_field: query_options.get(configs.user_script_env_field, [b'']),
-                configs.condor_pool: query_options.get(configs.condor_pool, [b'']),
-                configs.user_n_cores: query_options.get(configs.user_n_cores, [b'2']),
-                configs.user_memory: query_options.get(configs.user_memory, [b'8']),
-                configs.notebook: query_options.get(configs.notebook, [b'']),
-            }
-            return await self.post(user_name=for_user, server_name=server_name)
+        try:
+            await super().get(for_user, server_name)
+        except web.HTTPError as e:
+            form = await self._render_form_wrapper(user, message=e.message)
+            self.finish(form)
+            return
 
     @web.authenticated
     def post(self, user_name=None, server_name=''):
@@ -198,7 +179,6 @@ class SpawnHandler(JHSpawnHandler):
             query_params = {
                 "env": configs.env_name.format(project_folder=project_folder),
                 "repo": options.get(configs.repository),
-                configs.notebook: options.get(configs.notebook, ''),
             }
             if options.get(configs.customenv_type) == configs.accpy_special_type:
                 query_params[options.get(configs.customenv_type)] = options.get(configs.customenv_type_version)
@@ -210,7 +190,7 @@ class SpawnHandler(JHSpawnHandler):
         else:
             next_url = self.get_next_url(
                 user,
-                default=url_path_join(self.hub.base_url, "spawn-pending", user.escaped_name, server_name, options.get(configs.notebook, '')),
+                default=url_path_join(self.hub.base_url, "spawn-pending", user.escaped_name, server_name),
             )
         self.redirect(next_url)
 
--- a/SwanHub/swanhub/spawn_handler.py
+++ b/SwanHub/swanhub/spawn_handler.py
@@ -159,7 +159,6 @@ class SpawnHandler(JHSpawnHandler):
         if options.get(configs.source_type) == configs.customenv_special_type:
             # Add the query parameters to the URL
             query_params = {
-                "env": configs.env_name.format(project_folder=options.get(configs.project_folder)), # {reponame}_env or {lastfolder}_env
                 "repo": options.get(configs.repository),
             }
             if options.get(configs.builder) == configs.accpy_special_type:
--- a/SwanHub/swanhub/spawn_handler.py
+++ b/SwanHub/swanhub/spawn_handler.py
@@ -55,12 +55,33 @@ class SpawnHandler(JHSpawnHandler):
             self.finish(form)
             return
 
-        try:
-            await super().get(for_user, server_name)
-        except web.HTTPError as e:
-            form = await self._render_form_wrapper(user, message=e.message)
-            self.finish(form)
-            return
+        if not self.request.query_arguments:
+            try:
+                await super().get(for_user, server_name)
+            except web.HTTPError as e:
+                form = await self._render_form_wrapper(user, message=e.message)
+                self.finish(form)
+                return
+        else:
+            query_options = {}
+            for key, byte_list in self.request.query_arguments.items():
+                query_options[key] = [bs.decode('utf8') for bs in byte_list]
+            self.request.body_arguments.update({
+                configs.source_type: query_options.get(configs.source_type, configs.lcg_special_type).encode(),
+                configs.repository: query_options.get(configs.repository, '').encode(),
+                configs.repository_type: query_options.get(configs.repository_type, '').encode(),
+                configs.customenv_type: query_options.get(configs.customenv_type, '').encode(),
+                configs.customenv_type_version: query_options.get(configs.customenv_type_version, '').encode(),
+                configs.lcg_rel_field: query_options.get(configs.lcg_rel_field, '').encode(),
+                configs.platform_field: query_options.get(configs.platform_field, configs.default_platform).encode(),
+                configs.spark_cluster_field: query_options.get(configs.spark_cluster_field, '').encode(),
+                configs.user_script_env_field: query_options.get(configs.user_script_env_field, '').encode(),
+                configs.condor_pool: query_options.get(configs.condor_pool, '').encode(),
+                configs.user_n_cores: query_options.get(configs.user_n_cores, '2').encode(),
+                configs.user_memory: query_options.get(configs.user_memory, '8Gi').encode(),
+                configs.notebook: query_options.get(configs.notebook, '').encode(),
+            })
+            return await self.post(user_name=for_user, server_name=server_name)
 
     @web.authenticated
     def post(self, user_name=None, server_name=''):
@@ -111,16 +132,16 @@ class SpawnHandler(JHSpawnHandler):
         try:
             options = await maybe_future(spawner.run_options_from_form(form_options))
 
-            if options[configs.source_type] == configs.customenv_special_type and not options[configs.requirements]:
+            if options.get(configs.source_type) == configs.customenv_special_type and not options.get(configs.repository):
                 raise ValueError("Requirements not provided")
 
-            # Dont allow the session to spawn if the requirements are not valid
-            eos_match = re.match(configs.eos_pattern, options[configs.requirements])
-            git_match = re.match(configs.git_pattern, options[configs.requirements])
-            if options[configs.requirements_type] == configs.eos_special_type and not eos_match:
-                raise ValueError(f"Invalid EOS path for requirements: {options[configs.requirements]}")
-            if options[configs.requirements_type] == configs.git_special_type and not git_match:
-                raise ValueError(f"Invalid Git repository for requirements: {options[configs.requirements]}")
+            # Dont allow the session to spawn if the repository are not valid
+            eos_match = re.match(configs.eos_pattern, options.get(configs.repository))
+            git_match = re.match(configs.git_pattern, options.get(configs.repository))
+            if options.get(configs.repository_type) == configs.eos_special_type and not eos_match:
+                raise ValueError(f"Invalid EOS path for requirements: {options.get(configs.repository)}")
+            if options.get(configs.repository_type) == configs.git_special_type and not git_match:
+                raise ValueError(f"Invalid Git repository for requirements: {options.get(configs.repository)}")
 
             await self.spawn_single_user(user, server_name=server_name, options=options)
 
@@ -169,16 +190,19 @@ class SpawnHandler(JHSpawnHandler):
         if current_user is user:
             self.set_login_cookie(user)
 
-        if options[configs.source_type] == configs.customenv_special_type:
-            # Redirect to the customenvs page with the correct query parameters (/eos/user/.../requirements.txt)
-            project_folder = options[configs.requirements].split('/')[-2]
-            if options[configs.requirements].startswith('http'):
+        if options.get(configs.source_type) == configs.customenv_special_type:
+            project_folder =  options.get(configs.repository).split('/')[-1]
+            if options.get(configs.repository).startswith('http'):
                 project_folder = git_match[2] if git_match else user.escaped_name
             
             # Add the query parameters to the URL
-            query_params = {"env": configs.env_name.format(project_folder=project_folder), "req": options[configs.requirements]}
-            if options[configs.customenv_type] == configs.accpy_special_type:
-                query_params[options[configs.customenv_type]] = options[configs.customenv_type_version]
+            query_params = {
+                "env": configs.env_name.format(project_folder=project_folder),
+                "repo": options.get(configs.repository),
+                configs.notebook: options.get(configs.notebook, ''),
+            }
+            if options.get(configs.customenv_type) == configs.accpy_special_type:
+                query_params[options.get(configs.customenv_type)] = options.get(configs.customenv_type_version)
 
             next_url = self.get_next_url(
                 user,
@@ -187,7 +211,7 @@ class SpawnHandler(JHSpawnHandler):
         else:
             next_url = self.get_next_url(
                 user,
-                default=url_path_join(self.hub.base_url, "spawn-pending", user.escaped_name, server_name),
+                default=url_path_join(self.hub.base_url, "spawn-pending", user.escaped_name, server_name, options.get(configs.notebook, '')),
             )
         self.redirect(next_url)
 
@@ -243,7 +267,7 @@ class SpawnHandler(JHSpawnHandler):
                 metrics.append((metric, (date, 1)))
 
         spawn_context_key = ".".join(
-            [options[configs.lcg_rel_field], options[configs.spark_cluster_field]])
+            [options.get(configs.lcg_rel_field), options.get(configs.spark_cluster_field)])
         if not spawn_exception:
             # Add spawn success (no exception) and duration to the log and send as metrics
             spawn_exc_class = "None"
--- a/SwanHub/swanhub/spawn_handler.py
+++ b/SwanHub/swanhub/spawn_handler.py
@@ -115,10 +115,10 @@ class SpawnHandler(JHSpawnHandler):
                 raise ValueError("Requirements not provided")
 
             # Dont allow the session to spawn if the requirements are not valid
-            cernbox_match = re.match(configs.cernbox_pattern, options[configs.requirements])
+            eos_match = re.match(configs.eos_pattern, options[configs.requirements])
             git_match = re.match(configs.git_pattern, options[configs.requirements])
-            if options[configs.requirements_type] == configs.cernbox_special_type and not cernbox_match:
-                raise ValueError(f"Invalid CERNBox folder for requirements: {options[configs.requirements]}")
+            if options[configs.requirements_type] == configs.eos_special_type and not eos_match:
+                raise ValueError(f"Invalid EOS folder for requirements: {options[configs.requirements]}")
             if options[configs.requirements_type] == configs.git_special_type and not git_match:
                 raise ValueError(f"Invalid Git repository for requirements: {options[configs.requirements]}")
 
--- a/SwanHub/swanhub/spawn_handler.py
+++ b/SwanHub/swanhub/spawn_handler.py
@@ -118,7 +118,7 @@ class SpawnHandler(JHSpawnHandler):
             eos_match = re.match(configs.eos_pattern, options[configs.requirements])
             git_match = re.match(configs.git_pattern, options[configs.requirements])
             if options[configs.requirements_type] == configs.eos_special_type and not eos_match:
-                raise ValueError(f"Invalid EOS folder for requirements: {options[configs.requirements]}")
+                raise ValueError(f"Invalid EOS path for requirements: {options[configs.requirements]}")
             if options[configs.requirements_type] == configs.git_special_type and not git_match:
                 raise ValueError(f"Invalid Git repository for requirements: {options[configs.requirements]}")
 
@@ -170,8 +170,8 @@ class SpawnHandler(JHSpawnHandler):
             self.set_login_cookie(user)
 
         if options[configs.source_type] == configs.customenv_special_type:
-            # Redirect to the customenvs page with the correct query parameters
-            project_folder = options[configs.requirements].split('/').pop()
+            # Redirect to the customenvs page with the correct query parameters (/eos/user/.../requirements.txt)
+            project_folder = options[configs.requirements].split('/')[-2]
             if options[configs.requirements].startswith('http'):
                 project_folder = git_match[2] if git_match else user.escaped_name
             
--- a/SwanHub/swanhub/spawn_handler.py
+++ b/SwanHub/swanhub/spawn_handler.py
@@ -137,7 +137,6 @@ class SpawnHandler(JHSpawnHandler):
                     user, options, time.time() - start_time_spawn)
 
         except Exception as e:
-
             self._log_spawn_metrics(
                 user, options, time.time() - start_time_spawn, e)
 
@@ -231,7 +230,7 @@ class SpawnHandler(JHSpawnHandler):
                 metrics.append((metric, (date, 1)))
 
         spawn_context_key = ".".join(
-            [options.get(configs.lcg_rel_field, "CustomEnv"), options.get(configs.spark_cluster_field, "")])
+            [options.get(configs.lcg_rel_field, "CustomEnv"), options.get(configs.spark_cluster_field, "none")])
         if not spawn_exception:
             # Add spawn success (no exception) and duration to the log and send as metrics
             spawn_exc_class = "None"
--- a/SwanHub/swanhub/spawn_handler.py
+++ b/SwanHub/swanhub/spawn_handler.py
@@ -156,7 +156,7 @@ class SpawnHandler(JHSpawnHandler):
         if current_user is user:
             self.set_login_cookie(user)
 
-        if options.get(configs.source_type) == configs.customenv_special_type:
+        if options.get(configs.software_source) == configs.customenv_special_type:
             # Add the query parameters to the URL
             query_params = {
                 "repo": options.get(configs.repository),
--- a/SwanHub/swanhub/spawn_handler.py
+++ b/SwanHub/swanhub/spawn_handler.py
@@ -160,7 +160,7 @@ class SpawnHandler(JHSpawnHandler):
             # Add the query parameters to the URL
             query_params = {
                 "repo": options.get(configs.repository),
-                "repo_type": options.get(configs.repository_type),
+                "repo_type": options.get(configs.repo_type),
             }
             if options.get(configs.builder) == configs.accpy_special_type:
                 query_params[options.get(configs.builder)] = options.get(configs.builder_version)
--- a/SwanHub/swanhub/spawn_handler.py
+++ b/SwanHub/swanhub/spawn_handler.py
@@ -157,7 +157,7 @@ class SpawnHandler(JHSpawnHandler):
         if current_user is user:
             self.set_login_cookie(user)
 
-        if options.get(configs.source_type) == configs.customenv_special_type:            
+        if options.get(configs.source_type) == configs.customenv_special_type:
             # Add the query parameters to the URL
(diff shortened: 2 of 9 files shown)


SwanSpawner/swanspawner/templates/options_form_template.html
```diff
@@ -6,6 +6,25 @@
...
+      {
+        "value": "accpy-2021.12",
+        "text": "Acc-Py 2021.12"
+      },
+      {
+        "value": "accpy-2020.11",
+        "text": "Acc-Py 2020.11"
+      }
+    ];
+
```
etejedor, line 27: This looks to me like information that should be part of the options_form.json, right? In the same way we list all the available LCGs and their allowed capabilities in the latter. Can you check how you would integrate it there? Otherwise, every time we support new acc-py distributions we need to come here and change the template.

SwanSpawner/swanspawner/templates/options_form_template.html
```diff
@@ -26,6 +45,32 @@
...
+        }
+    }
+
+    /*
+    * Selects default LCG release (105)
+    */
+   function selectDefaultLCG(id) {
+        var selector = document.getElementById(id);
+        for (var i = 0; i < selector.options.length; i++) {
+            if (selector.options[i].value === 'LCG_105_swan') {
```
etejedor, line 67: This is not anymore the default LCG release, and we can't hardcode any LCG release here otherwise we would need to modify this template every time a new LCG stack is released. Can you just take the one at the top of the options_form_config.json? Or alternatively add some attribute to an LCG release entry in the options_form_config.json that marks it as default.

SwanSpawner/swanspawner/templates/options_form_template.html
```diff
@@ -170,6 +215,77 @@
...
+        
+        document.body.appendChild(tempElement);
+        const width = tempElement.offsetWidth;
+        document.body.removeChild(tempElement);
+        
+        selectElement.style.width = `${width + 50}px`;
+
+        const requirementsOption = document.getElementById('requirementsOption');
+        if (selectElement.value === 'eos') {
+            requirementsOption.placeholder = 'e.g. $CERNBOX_HOME/MySWAN';
```
etejedor, line 257:
```suggestion
            requirementsOption.placeholder = 'e.g. $CERNBOX_HOME/MyFolder';
```

SwanSpawner/swanspawner/templates/options_form_template.html
```diff
@@ -170,6 +215,77 @@
...
+        } else if (selectElement.value === 'git') {
+            requirementsOption.placeholder= "e.g. https://gitlab.cern.ch/user/myrepo";
+        }
+    }
+
+    /**
+     * Modifies the some elements according to the selected type of configuration (LCG or Custom Environments)
```
etejedor, line 264:
```suggestion
     * Modifies the displayed form according to the selected type of configuration (LCG or Custom Environments)
```

SwanSpawner/swanspawner/templates/options_form_template.html
```diff
@@ -170,6 +215,77 @@
...
+        var external_config = document.getElementById('external_config');
+        var software_config = document.getElementById('software_config');
+        
+        if (lcgOption.checked) {
+            software_config.style.display = 'block';
+            customenv_config.style.display = 'none';
+            external_config.style.display = 'block';
+            adjust_form();
+            adjust_spark();
+        } else if (customenvOption.checked) {
```
etejedor, line 280: One of the two options needs to be checked necessarily, no? Then why else if ?

SwanSpawner/swanspawner/templates/options_form_template.html
```diff
@@ -170,6 +215,77 @@
...
+
+            selectCustomenv_type.add(customenv_typeOption);
+        }
+    }
+
+    /**
+     * Allows to adjust the width of the dropdown element based on the selected option
+     */
+    function adjustDropdownWidth() {
+        const selectElement = document.getElementById('requirements_dropdown');
```
etejedor, line 241: `requirements_dropdown` should be renamed to `repository_type_dropdown` perhaps? I understand this is the little dropdown that shows EOS or git

SwanSpawner/swanspawner/templates/options_form_template.html
```diff
@@ -170,6 +215,77 @@
...
+            }
+
+            selectCustomenv_type.add(customenv_typeOption);
+        }
+    }
+
+    /**
+     * Allows to adjust the width of the dropdown element based on the selected option
+     */
+    function adjustDropdownWidth() {
```
etejedor, line 240: If this is just for the repository dropdown, the function name should include it (it's not a generic function). I.e. `function adjustRepositoryDropdown` would be enough.

SwanSpawner/swanspawner/templates/options_form_template.html
```diff
@@ -170,6 +215,77 @@
...
+            if (i === 0) {
+                customenv_typeOption.selected = true;
+            }
+
+            selectCustomenv_type.add(customenv_typeOption);
+        }
+    }
+
+    /**
+     * Allows to adjust the width of the dropdown element based on the selected option
```
etejedor, line 238: If this is a function for the repository type dropdown, the doc of the function needs to mention it explicitly (it's not generic). I.e. `Adjusts the width of the repository type dropdown element based on the selected option`

SwanSpawner/swanspawner/templates/options_form_template.html
```diff
@@ -170,6 +215,77 @@
...
+        tempElement.style.fontSize = window.getComputedStyle(selectElement).fontSize;
+        tempElement.innerHTML = selectedOptionText;
+        
+        document.body.appendChild(tempElement);
+        const width = tempElement.offsetWidth;
+        document.body.removeChild(tempElement);
+        
+        selectElement.style.width = `${width + 50}px`;
+
+        const requirementsOption = document.getElementById('requirementsOption');
```
etejedor, line 255:
```suggestion
        const requirementsOption = document.getElementById('repositoryOption');
```

SwanSpawner/swanspawner/templates/options_form_template.html
```diff
@@ -170,6 +215,77 @@
...
+            requirementsOption.placeholder = 'e.g. $CERNBOX_HOME/MySWAN';
+        } else if (selectElement.value === 'git') {
+            requirementsOption.placeholder= "e.g. https://gitlab.cern.ch/user/myrepo";
+        }
+    }
+
+    /**
+     * Modifies the some elements according to the selected type of configuration (LCG or Custom Environments)
+     */
+    function toggle_forms() {
```
etejedor, line 266:
```suggestion
    function toggle_form() {
```

SwanSpawner/swanspawner/templates/options_form_template.html
```diff
@@ -170,6 +215,77 @@
...
+        var software_config = document.getElementById('software_config');
+        
+        if (lcgOption.checked) {
+            software_config.style.display = 'block';
+            customenv_config.style.display = 'none';
+            external_config.style.display = 'block';
+            adjust_form();
+            adjust_spark();
+        } else if (customenvOption.checked) {
+            selectDefaultLCG('lcgReleaseOptions');
```
etejedor, line 281: Why do we need to select the default LCG here?

SwanSpawner/swanspawner/templates/options_form_template.html
RodrigoSobral2000: Yeah, I've been changing a few things meanwhile, including that. I just got the first option from the recommended section

SwanSpawner/swanspawner/templates/options_form_template.html
RodrigoSobral2000: Yup, I just used else if in case we have more in the future

SwanSpawner/swanspawner/templates/options_form_template.html
RodrigoSobral2000: I was getting some bugs when I start switching the lcg options while changing between customenvs and lcg forms. From what I read from the adjust_form function, that was the way I found to make it work properly.

SwanSpawner/swanspawner/templates/options_form_template.html
RodrigoSobral2000: Yes, I'm just finishing a few details, I'll updated as soon as it gets stable

SwanSpawner/swanspawner/templates/options_form_template.html
RodrigoSobral2000: Yeah, I thought about that and actually it was something that I wanted to tell you about. This is definitely a temporary solution. I'll get a way to redo it soon

RodrigoSobral2000: We gotta a new PR in gitops to increment these options to QA in the same way is done for LCG

SwanSpawner/swanspawner/swanspawner.py
```diff
@@ -76,19 +105,50 @@ def __init__(self, **kwargs):
...
+            options[self.platform_field]        = platform
             options[self.user_script_env_field] = formdata[self.user_script_env_field][0]
             options[self.spark_cluster_field]   = formdata[self.spark_cluster_field][0] if self.spark_cluster_field in formdata.keys() else 'none'
             options[self.condor_pool]           = formdata[self.condor_pool][0]
             options[self.user_n_cores]          = int(formdata[self.user_n_cores][0])
             options[self.user_memory]           = formdata[self.user_memory][0] + 'G'
 
             self.offload = options[self.spark_cluster_field] != 'none'
-
+
```
etejedor, line 151: Remove spacing.

SwanSpawner/swanspawner/templates/options_form_template.html
```diff
@@ -170,6 +226,61 @@
 
             selectCondor.add(selectCondorOption);
         }
+
+    }
+
+    /**
+     * Adjusts the width of the repository type dropdown element based on the selected option
+     */
+    function adjustRepositoryDropdown() {
```
etejedor, line 235: I just noticed that some of the names of the functions you are adding are in camel case, like this one. We should stick to the snake case style that was in the file before for our own functions.

SwanSpawner/swanspawner/templates/options_form_template.html
```diff
@@ -170,6 +226,61 @@
...
+        var software_config = document.getElementById('software_config');
+        
+        if (lcgOption.checked) {
+            software_config.style.display = 'block';
+            customenv_config.style.display = 'none';
+            external_config.style.display = 'block';
+            adjust_form();
+            adjust_spark();
+        } else { // if customenvOption.checked
+            selectDefaultLCG();
```
etejedor, line 276: Why do we need to select default lcg in the case of custom env?

SwanSpawner/swanspawner/templates/options_form_template.html
```diff
@@ -204,54 +314,131 @@
...
-    <br>
-    <label for="scriptenvOption">Environment script <a href="#" onclick="toggle_visibility('scriptenvDetails');"><span class='nbs'>more...</span></a>
-    <div style="display:none;" id="scriptenvDetails">
-       <span class='nb'>User-provided bash script to define custom environment variables. The variable CERNBOX_HOME is resolved to the proper /eos/user/u/username directory.</span>
+
+    <div id="source_typeSection">
+        <h2>Software</h2>
+        <label>Source <a href="#" onclick="toggle_visibility('sourceDetails');"><span class='nbs'>more...</span></a>
+            <div style="display:none;" id="sourceDetails">
+                <span class='nb'>Source type for your container.</span>
```
etejedor, line 322:
```suggestion
                <span class='nb'>Software source: curated stack (LCG) or custom environment.</span>
```

SwanSpawner/swanspawner/templates/options_form_template.html
```diff
@@ -204,54 +314,131 @@
...
+                <div style="display:none;" id="lcgReleaseDetails">
+                    <span class='nb'>Software stack to associate to the container. See the <a target="_blank" href="http://lcginfo.cern.ch/">LCG package info</a> page.</span>
+                </div>
+            </label>
+            <select id="lcgReleaseOptions" name="LCG-rel" onchange="adjust_form();">
+            </select>
+        </div>
+        <br>
+        
+        <div id="plataformSection">
```
etejedor, line 345:
```suggestion
        <div id="platformSection">
```

SwanSpawner/swanspawner/templates/options_form_template.html
```diff
@@ -204,54 +314,131 @@
...
-    <br>
-    <label for="coresOptions">Number of cores <a href="#" onclick="toggle_visibility('ncoresDetails');"><span class='nbs'>more...</span></a>
-    <div style="display:none;" id="ncoresDetails">
-       <span class='nb'>Number of cores to associate to the container.</span>
+    
+    <div id="software_config">
+        <div id="lcgSection">
+            <label for="lcgReleaseOptions">Software stack <a href="#" onclick="toggle_visibility('lcgReleaseDetails');"><span class='nbs'>more...</span></a>
+                <div style="display:none;" id="lcgReleaseDetails">
+                    <span class='nb'>Software stack to associate to the container. See the <a target="_blank" href="http://lcginfo.cern.ch/">LCG package info</a> page.</span>
```
etejedor, line 337:
```suggestion
                    <span class='nb'>LCG release to use as software source. See the <a target="_blank" href="http://lcginfo.cern.ch/">LCG package info</a> page.</span>
```

SwanSpawner/swanspawner/templates/options_form_template.html
```diff
@@ -204,54 +314,131 @@
...
+            </label>
+            <select id="lcgReleaseOptions" name="LCG-rel" onchange="adjust_form();">
+            </select>
+        </div>
+        <br>
+        
+        <div id="plataformSection">
+            <label for="platformOptions">Platform <a href="#" onclick="toggle_visibility('platformDetails');"><span class='nbs'>more...</span></a>
+                <div style="display:none;" id="platformDetails">
+                    <span class='nb'>Combination of compiler version and flags.</span>
```
etejedor, line 348:
```suggestion
                    <span class='nb'>Operating system and compiler version.</span>
```

SwanSpawner/swanspawner/templates/options_form_template.html
```diff
@@ -204,54 +314,131 @@
...
     </div>
-    </label>
-    <input type="text" id="scriptenvOption" name="scriptenv" placeholder="e.g. $CERNBOX_HOME/MySWAN/myscript.sh">
-    <br>
-    <label for="coresOptions">Number of cores <a href="#" onclick="toggle_visibility('ncoresDetails');"><span class='nbs'>more...</span></a>
-    <div style="display:none;" id="ncoresDetails">
-       <span class='nb'>Number of cores to associate to the container.</span>
+    
+    <div id="software_config">
+        <div id="lcgSection">
```
etejedor, line 334:
```suggestion
        <div id="lcgReleaseSection">
```

SwanSpawner/swanspawner/templates/options_form_template.html
```diff
@@ -204,54 +314,131 @@
...
+        <br>
     </div>
-    </label>
-    <select id="coresOptions" name="ncores"></select>
-    <br>
-    <label for="memoryOptions">Memory <a href="#" onclick="toggle_visibility('memoryDetails');"><span class='nbs'>more...</span></a>
-    <div style="display:none;" id="memoryDetails">
-       <span class='nb'>Amount of Memory allocated to the container.</span>
+
+    <!-- Custom environment configuration -->
```
etejedor, line 366: If you add a comment here to mark the start of the custom env options (which is legit), you should do the same above with the start of the lcg options.

SwanSpawner/swanspawner/templates/options_form_template.html
```diff
@@ -204,54 +314,131 @@
...
+    <div id="customenv_config" style="display: none;">
+        <div id="repositorySection">
+            <label for="repositoryOption">Repository 
+                <a href="#" onclick="toggle_visibility('repositoryDetails');"><span class='nbs'>more...</span></a>
+                <div style="display:none;" id="repositoryDetails">
+                    <span class='nb'>Path to a folder stored in your <a target="_blank" href="https://cernbox.cern.ch/">EOS</a> (/eos/user/u/username), or a <a target="_blank" href="https://github.com/new/">GitHub</a>/<a target="_blank" href="https://gitlab.cern.ch/projects/new">CERN GitLab</a> repository containing a requirements.txt in its root path.</span>
```
etejedor, line 372:
```suggestion
                    <span class='nb'>Repository containing a requirements.txt file. It can be a path on EOS or the URL of a public git repository.</span>
```

SwanSpawner/swanspawner/templates/options_form_template.html
RodrigoSobral2000: I think with the changes I've been made that became obsolete, so I'll removed it and check if its fine

SwanSpawner/swanspawner/templates/options_form_template.html
```diff
@@ -204,54 +314,131 @@
...
+                </select>
+                <input type="text" id="repositoryOption" name="repository" style="flex-grow: 1; margin-left: 0; border: 1px solid #afafaf;">
+            </div>
+        </div>
+        <br>
+        
+        <div id="customenv_typeSection">
+            <label for="customenv_typeOptions">Type <a href="#" onclick="toggle_visibility('customenv_typeDetails');"><span class='nbs'>more...</span></a>
+                <div style="display:none;" id="customenv_typeDetails">
+                    <span class='nb'>Type of custom environment to create. See more in the <a target="_blank" href="https://confluence.cern.ch/display/ACCPY/Getting+started+with+Acc-Py">Acc-Py User guide</a></span>
```
etejedor, line 389:
```suggestion
                    <span class='nb'>Type of custom environment to create.</a></span>
```

etejedor: I know it's a simplification, but I found the information in the link you specify misleading, as they won't need to run those instructions from SWAN.

SwanHub/swanhub/handlers_configs.py
```diff
@@ -15,12 +15,84 @@ class SpawnHandlersConfigs(SingletonConfigurable):
...
+    user_n_cores = 'ncores'
+
+    user_memory = 'memory'
+
+    spark_cluster_field = 'spark-cluster'
+
+    condor_pool = 'condor-pool'
+
+    eos_pattern = Unicode(
+        default_value=r'^(?:\$CERNBOX|(?:/[^/\n]+)*/[^/\n]+)?$',
```
etejedor, line 43: If you are matching an EOS pattern I would expect the pattern to start with `/eos`?
Watch out, it doesn't need to be just CERNBOX (i.e. /eos/user/u/username), it can be e.g. /eos/cms or /eos/projects or...

etejedor: Also, does this need to be configurable?

SwanHub/swanhub/handlers_configs.py
```diff
@@ -15,12 +15,84 @@ class SpawnHandlersConfigs(SingletonConfigurable):
...
+        config=True,
+        help='Regular expression pattern for the repository provided by a EOS folder.'
+    )
+
+    eos_special_type = Unicode(
```
etejedor, line 48: Does this need to be configurable?

SwanHub/swanhub/handlers_configs.py
etejedor: Ok, we might want to make it configurable, yes...

SwanHub/swanhub/handlers_configs.py
```diff
@@ -15,12 +15,84 @@ class SpawnHandlersConfigs(SingletonConfigurable):
...
+        help='Special type for the repository provided by a EOS folder.'
+    )
+
+    git_pattern = Unicode(
+        default_value=r'https?://(?:github\.com|gitlab\.cern\.ch)/([^/\s]+)/([^/\s]+)/?',
+        config=True,
+        help='Regular expression pattern for the repository provided by a GitLab or GitHub repository.'
+    )
+
+    git_special_type = Unicode(
```
etejedor, line 60: Does this need to be configurable?

SwanHub/swanhub/handlers_configs.py
```diff
@@ -15,12 +15,84 @@ class SpawnHandlersConfigs(SingletonConfigurable):
...
+        default_value='git',
+        config=True,
+        help='Special repository type for Git repositories.'
+    )
+
+    lcg_special_type = Unicode(
```
etejedor, line 66: Does this need to be configurable?

SwanHub/swanhub/handlers_configs.py
```diff
@@ -15,12 +15,84 @@ class SpawnHandlersConfigs(SingletonConfigurable):
...
+        default_value='lcg',
+        config=True,
+        help='Special type for default sourcing.'
+    )
+
+    customenv_special_type = Unicode(
```
etejedor, line 72: Does this need to be configurable?

SwanHub/swanhub/handlers_configs.py
```diff
@@ -15,12 +15,84 @@ class SpawnHandlersConfigs(SingletonConfigurable):
...
+        help='Special type for sourcing the environment.'
+    )
+
+    accpy_special_type = Unicode(
+        default_value='accpy',
+        config=True,
+        help='Special type for custom environments.'
+    )
+
+    env_name = Unicode(
```
etejedor, line 84: So the name we'll pick for the environment is the name of the git repo or the last folder of the EOS path + _env?

SwanHub/swanhub/handlers_configs.py
RodrigoSobral2000: Does this mean that the only restriction on the path is the fact that it starts with /eos?

SwanSpawner/swanspawner/templates/options_form_template.html
```diff
@@ -204,54 +314,131 @@
...
+            <label for="repositoryOption">Repository 
+                <a href="#" onclick="toggle_visibility('repositoryDetails');"><span class='nbs'>more...</span></a>
+                <div style="display:none;" id="repositoryDetails">
+                    <span class='nb'>Path to a folder stored in your <a target="_blank" href="https://cernbox.cern.ch/">EOS</a> (/eos/user/u/username), or a <a target="_blank" href="https://github.com/new/">GitHub</a>/<a target="_blank" href="https://gitlab.cern.ch/projects/new">CERN GitLab</a> repository containing a requirements.txt in its root path.</span>
+                </div>
+            </label>
+            <br>
+            <div style="display: flex;">
+                <select id="repository_type_dropdown" name="repository_type" onchange="adjustRepositoryDropdown();" style="width: auto; display: inline-flex; border: 1px solid #afafaf; background-color: #efefef;">
+                      <option value="git">Git</option>
```
etejedor, line 378: The default is git, right?

SwanHub/swanhub/handlers_configs.py
RodrigoSobral2000: Yup, that's it. What do you think?

SwanHub/swanhub/handlers_configs.py
etejedor: `/eos` + a finite number of folder names separated by slashes

SwanHub/swanhub/handlers_configs.py
etejedor: Also, why `special`? We can perhaps make all these types shorter without the special adjective?

SwanSpawner/swanspawner/templates/options_form_template.html
RodrigoSobral2000: Yes, it is the first option

SwanHub/swanhub/spawn_handler.py
```diff
@@ -109,6 +111,18 @@ async def _post(self, user_name, server_name):
 
         try:
             options = await maybe_future(spawner.run_options_from_form(form_options))
+
+            if options.get(configs.source_type) == configs.customenv_special_type and not options.get(configs.repository):
+                raise ValueError("Requirements not provided")
```
etejedor, line 116:
```suggestion
                raise ValueError("Repository not provided")
```

etejedor: Question: can't we make certain form items to be required in the options form template?

SwanHub/swanhub/spawn_handler.py
```diff
@@ -109,6 +111,18 @@ async def _post(self, user_name, server_name):
...
+
+            # Dont allow the session to spawn if the repository are not valid
```
etejedor, line 118:
```suggestion
            # Do not allow the session to spawn if the repository is not valid
```

SwanHub/swanhub/handlers_configs.py
RodrigoSobral2000: Ok, I thought we would only work the this structure /eos/user/u/username
I'll give it a fix then

SwanHub/swanhub/spawn_handler.py
RodrigoSobral2000: Yes, but I played safe once again because it could also complain when launching a normal lcg environment. I will test that to make sure

SwanHub/swanhub/spawn_handler.py
```diff
@@ -109,6 +111,18 @@ async def _post(self, user_name, server_name):
...
+            eos_match = re.match(configs.eos_pattern, options.get(configs.repository))
```
etejedor, line 119: Optimization: you only try to match a repo type if the user has selected that type, so we need an if else here with the matches inside the correspoding if/else block. Match operations are relatively costly, so better to save on them.

SwanHub/swanhub/spawn_handler.py
```diff
@@ -155,12 +169,29 @@ def _finish_spawn(f):
...
+                "repo": options.get(configs.repository),
+            }
+            if options.get(configs.customenv_type) == configs.accpy_special_type:
+                query_params[options.get(configs.customenv_type)] = options.get(configs.customenv_type_version)
+
+            next_url = self.get_next_url(
+                user,
+                default=url_concat(url_path_join("user", user.escaped_name, "customenvs", server_name), query_params),
+            )
+        else:
```
etejedor, line 190:
```suggestion
        else:  # LCG release
```

SwanHub/swanhub/spawn_handler.py
```diff
@@ -155,12 +169,29 @@ def _finish_spawn(f):
...
-            user,
-            default=url_path_join(
-                self.hub.base_url, "spawn-pending", user.escaped_name, server_name
-            ),
-        )
+
+        if options.get(configs.source_type) == configs.customenv_special_type:
+            project_folder =  options.get(configs.repository).split('/')[-1]
+            if options.get(configs.repository).startswith('http'):
+                project_folder = git_match[2] if git_match else user.escaped_name
```
etejedor, line 176: This line deserves a comment?

SwanHub/swanhub/spawn_handler.py
RodrigoSobral2000: Yeah, it starts the session either way. From what I read, the required inputs gotta be within a explicit <form> tag to make it work

SwanHub/swanhub/spawn_handler.py
```diff
@@ -155,12 +169,29 @@ def _finish_spawn(f):
...
+            
+            # Add the query parameters to the URL
+            query_params = {
+                "env": configs.env_name.format(project_folder=project_folder),
+                "repo": options.get(configs.repository),
```
etejedor, line 181: I guess if we do a good pattern matching with the repo name we do not need to validate it before sending it in the query? Meaning there's no malicious injection in it that could cause troubles in the backend (i.e. makenv script).

SwanHub/swanhub/spawn_handler.py
```diff
@@ -155,12 +169,29 @@ def _finish_spawn(f):
...
+            }
+            if options.get(configs.customenv_type) == configs.accpy_special_type:
+                query_params[options.get(configs.customenv_type)] = options.get(configs.customenv_type_version)
+
+            next_url = self.get_next_url(
```
etejedor, line 186: This deserves a comment, no? Like `# Trigger the execution of the SwanCustomEnvs extension with the corresponding query arguments`.

SwanHub/swanhub/spawn_handler.py
```diff
@@ -215,7 +246,7 @@ def _log_spawn_metrics(self, user, options, spawn_duration_sec, spawn_exception=
                 metrics.append((metric, (date, 1)))
 
         spawn_context_key = ".".join(
-            [options[configs.lcg_rel_field], options[configs.spark_cluster_field]])
+            [options.get(configs.lcg_rel_field), options.get(configs.spark_cluster_field)])
```
etejedor, line 249: Why this change? Because we do not necessarily get the lcg release and the spark cluster field values anymore? So we'll have two None here.
This is important because the log lines that are being built here are then parsed to build metrics. It's ok if the spark cluster field is None but, if the lcg_rel_field is None, can you put something like `CustomEnv` instead of that None? Otherwise session spawning with CustomEnv will just be logged with a None as LCG release, which is not too nice.

SwanSpawner/swanspawner/swandockerspawner.py
```diff
@@ -152,11 +152,11 @@ async def start(self):
             options_form_config_data = json.load(json_file)
 
         username = self.user.name
-        platform = self.user_options.get(self.platform_field,options_form_config_data["options"][1]['platforms'][0]['value'])
-        lcg_rel = self.user_options.get(self.lcg_rel_field,options_form_config_data["options"][1]["lcg"]["value"])
-        cluster = self.user_options.get(self.spark_cluster_field,options_form_config_data["options"][1]['clusters'][0]['value'])
-        cpu_quota = self.user_options.get(self.user_n_cores,int(options_form_config_data["options"][1]['cores'][0]['value']))
-        mem_limit = self.user_options.get(self.user_memory,options_form_config_data["options"][1]['memory'][0]['value'] + 'G')
+        platform = self.user_options.get(self.platform_field,options_form_config_data["lcg_options"][1]['platforms'][0]['value'])
```
etejedor, line 155: Nice catch! It was also here :)

SwanSpawner/swanspawner/swanspawner.py
```diff
@@ -41,6 +46,30 @@ class SwanSpawner(base_class):
 
         condor_pool = 'condor-pool'
 
+        customenv_special_type = Unicode(
+            default_value='customenv',
+            config=True,
+            help='Special type for custom environments.'
+        )
+
+        eos_special_type = Unicode(
```
etejedor, line 55: This is not used below.

SwanSpawner/swanspawner/swanspawner.py
```diff
@@ -41,6 +46,30 @@ class SwanSpawner(base_class):
...
+            default_value='eos',
+            config=True,
+            help='Special type for repository provided by a EOS folder.'
+        )
+
+        git_special_type = Unicode(
```
etejedor, line 61: This is not used below.

SwanHub/swanhub/handlers_configs.py
```diff
@@ -15,12 +15,84 @@ class SpawnHandlersConfigs(SingletonConfigurable):
...
+    repository = 'repository'
+
     lcg_rel_field = 'LCG-rel'
 
-    spark_cluster_field = 'spark-cluster'
+    platform_field = 'platform'
 
     user_script_env_field = 'scriptenv'
 
+    user_n_cores = 'ncores'
```
etejedor, line 34: Why did you add this?

SwanHub/swanhub/handlers_configs.py
```diff
@@ -15,12 +15,84 @@ class SpawnHandlersConfigs(SingletonConfigurable):
...
+
+    user_memory = 'memory'
+
+    spark_cluster_field = 'spark-cluster'
+
+    condor_pool = 'condor-pool'
```
etejedor, line 40: Why did you add this?

SwanHub/swanhub/handlers_configs.py
```diff
@@ -15,12 +15,84 @@ class SpawnHandlersConfigs(SingletonConfigurable):
...
     lcg_rel_field = 'LCG-rel'
 
-    spark_cluster_field = 'spark-cluster'
+    platform_field = 'platform'
 
     user_script_env_field = 'scriptenv'
 
+    user_n_cores = 'ncores'
+
+    user_memory = 'memory'
```
etejedor, line 36: Why did you add this?

SwanHub/swanhub/handlers_configs.py
```diff
@@ -15,12 +15,84 @@ class SpawnHandlersConfigs(SingletonConfigurable):
         Singleton class where all the configurations are stored
     """
 
+    source_type = 'source_type'
+
+    customenv_type = 'customenv_type'
+    
+    customenv_type_version = 'customenv_type_version'
```
etejedor, line 22:
```suggestion
    accpy_version = 'accpy_version'
```

etejedor: I understand this is what you actually mean with this variable.

SwanSpawner/swanspawner/swanspawner.py
```diff
@@ -41,6 +46,30 @@ class SwanSpawner(base_class):
...
+            default_value='git',
+            config=True,
+            help='Special repository type for Git repositories.'
+        )
+
+        default_platform = Unicode(
```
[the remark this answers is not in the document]
RodrigoSobral2000: Lets suppose the user starts selection a random lcg that has not Alma9 (e.g. lcg 101). If he then changes to customenv configuration, the platform in the form will be something else that isnt Alma9, right?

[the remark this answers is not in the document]
RodrigoSobral2000: I think we really need to make sure that we specify a standard platform for running the customenvs. Unless its optional and that's done elsewhere, not sure about that tbh

SwanSpawner/swanspawner/swanspawner.py
```diff
@@ -76,19 +105,50 @@ def __init__(self, **kwargs):
             if not self.options_form and self.options_form_config:
                 # if options_form not provided, use templated options form based on configuration file
                 self.options_form = self._render_templated_options_form
+        
+        def replace_eos_home(self, repo_path: str):
```
etejedor, line 109: `replace_cernbox_home`? Also, small docstring please.

SwanSpawner/swanspawner/swanspawner.py
```diff
@@ -136,6 +193,15 @@ def get_env(self):
             if self.user_options[self.condor_pool] != 'none':
                 env['CERN_HTCONDOR'] = 'true'
 
+            # Enable configuration for LCG and custom environments
+            if self.user_options[self.source_type] == self.customenv_special_type:
+                env['AUTOENV'] = "true"
```
etejedor, line 198: Why do we need this variable?

SwanSpawner/swanspawner/swanspawner.py
```diff
@@ -109,10 +169,7 @@ def get_env(self):
             if self.lcg_rel_field in self.user_options:
                 # session spawned via the form
                 env.update(dict(
-                    ROOT_LCG_VIEW_NAME     = self.user_options[self.lcg_rel_field],
-                    ROOT_LCG_VIEW_PLATFORM = self.user_options[self.platform_field],
-                    USER_ENV_SCRIPT        = self.user_options[self.user_script_env_field],
-                    ROOT_LCG_VIEW_PATH     = self.lcg_view_path,
+                    SOURCE_TYPE            = self.user_options[self.source_type],
```
etejedor, line 172: Do we need this variable?

SwanHub/swanhub/handlers_configs.py
RodrigoSobral2000:
```
if options.get(configs.customenv_type) == configs.accpy_special_type:
    query_params[options.get(configs.customenv_type)] = options.get(configs.customenv_type_version)
```

This is the only code that is used. I've declared the customenv_type_version variable so that it doesn't become hardcoded.
What happens here is adding a query argument like `&accpy=2023.06`, thats why I need to have the value `accpy-2023.06` in the options_form

SwanSpawner/swanspawner/swanspawner.py
```diff
@@ -109,10 +169,7 @@ def get_env(self):
             if self.lcg_rel_field in self.user_options:
```
etejedor, line 169: This condition should not be used anymore to decide whether to set these variables, right? We need them in the case of custom env as well?

SwanSpawner/swanspawner/swanspawner.py
```diff
@@ -76,19 +105,50 @@ def __init__(self, **kwargs):
...
+        
+        def replace_eos_home(self, repo_path: str):            
+            eos_path = self.eos_path_format.format(username=self.user.name)
+            return repo_path.replace('$CERNBOX_HOME', eos_path.rstrip('/'))
 
         def options_from_form(self, formdata):
+            source_type = formdata[self.source_type][0]
+            lcg = formdata[self.lcg_rel_field][0]
+            platform = formdata[self.platform_field][0]
+            aux_req = formdata[self.customenv_type][0].lower().split('-')
```
etejedor, line 117: What do you get here in `aux_req`? Is it the git/eos part or the generic/acc-pyXXXX part?

SwanSpawner/swanspawner/swanspawner.py
```diff
@@ -76,19 +105,50 @@ def __init__(self, **kwargs):
...
+            customenv_type, customenv_type_version = '', ''
+            if len(aux_req) == 2:
+                customenv_type, customenv_type_version = aux_req
+            repository, repository_type = '', ''
+            if source_type == self.customenv_special_type:
+                lcg, platform = '', self.default_platform
+                repository = formdata[self.repository][0]
+                repository_type = formdata[self.repository_type][0]
+                if repository.startswith("http"):
```
etejedor, line 126: Why do we do a match for the repository also here? I think we should just do a pattern match of the repository string in one place. I understand this is the first one to execute, so perhaps do all the matching here so you do not need to do it later in the spawn handler?

SwanHub/swanhub/handlers_configs.py
etejedor: Yes, but wouldn't `customenv_type_version` be better called `accpy_version`, if that's that it points too? I'm just trying to get shorter and more meaningful names here...

SwanSpawner/swanspawner/swanspawner.py
RodrigoSobral2000: aux_req isnt the right name, indeed. But the idea is to split the customenv_type (accpy-2023.06) into customenv_type (accpy) and customenv_type_version (2023.06)

SwanHub/swanhub/handlers_configs.py
etejedor: No, when we discussed it together with you and Diogo in the meeting we agreed to give full freedom and allow other paths on EOS to be specified.

SwanHub/swanhub/handlers_configs.py
etejedor: Yes, it is either that or username_env. I'm fine with any.

SwanSpawner/swanspawner/swanspawner.py
RodrigoSobral2000: This match is used for cleaning the repo url, getting rid of not needed stuff and not allowing to fail when cloning the repo

SwanHub/swanhub/handlers_configs.py
```diff
@@ -15,12 +15,84 @@ class SpawnHandlersConfigs(SingletonConfigurable):
...
+        help='Special type for custom environments.'
+    )
+
+    env_name = Unicode(
+        default_value='{project_folder}_env',
+        config=True,
+        help='Name format for custom environment launched by the user.'
+    )
+
+    default_platform = Unicode(
```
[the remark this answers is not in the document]
etejedor: But the platform will always come from the form, there is always an option that needs to be selected (it is not a free text box that can be empty). And also, in the case of custom envs there is no platform, since the platform is related to the OS and compiler that a particular LCG release was built for.

SwanSpawner/swanspawner/swanspawner.py
RodrigoSobral2000: Yes, its used in 02_environment to know if its supposed to load the lcg

SwanHub/swanhub/spawn_handler.py
etejedor: If it is easy to do, it will be easier to impose constraints before sending the form, it is more immediate, otherwise the request has to come here, we build the error message and then we return to the user. But admittedly I don't know how that is implemented in html.

etejedor: @diocas might want to kick in?

SwanSpawner/swanspawner/swanspawner.py
RodrigoSobral2000: To makenv know if its supposed to source the environment

SwanSpawner/swanspawner/swanspawner.py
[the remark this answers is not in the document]
etejedor: But the custom envs do not need any platform, right? The platform is just to know which build of the LCG release to take (now centos7 or alma9), together with a compiler version choice. For custom envs, we will always pick an alma9 image, but that is not what the platform field of the LCG release specifies.

Bottom line: customenvs do not need platform. LCG releases do, but you'll always get one from the form.

SwanSpawner/swanspawner/swanspawner.py
RodrigoSobral2000: Supposelly they're needed in case of an API call. So I think would be better to verify the opposite:
`
if self.lcg_rel_field not in self.user_options and self.source_type != self.customenv_special_type:
    # session spawned via the API
else:
    # session spawned via the form
`

SwanSpawner/swanspawner/templates/options_form_template.html
etejedor: I think it shouldn't be needed, when you toggle to Custom Env there shouldn't be any platform to choose in the form. And if you go back to LCG release, well you just display the platforms available for the selected LCG as always.

SwanHub/swanhub/spawn_handler.py
```diff
@@ -107,8 +107,13 @@ async def _post(self, user_name, server_name):
 
         start_time_spawn = time.time()
 
+        options = {}
         try:
             options = await maybe_future(spawner.run_options_from_form(form_options))
+
+            if options.get(configs.source_type) == configs.customenv_special_type and not options.get(configs.repository):
```
etejedor, line 114: Why is this validation not done in the spawner like the other ones?

RodrigoSobral2000: You're right, now that we fixed that issue about options, that statement doesn't need to be there

SwanHub/swanhub/handlers_configs.py
RodrigoSobral2000: I've been changing the name of customenv_type and customenv_type_version to builder and builder_version.
I think it makes much more sense, considering the options we deliver represent the software we have to build the custom environment the user wants. Not forgetting that it can be accpy, conda, experiments, etc. So, we don't keep our variable name with such a narrow scope.

SwanHub/swanhub/spawn_handler.py
```diff
@@ -215,7 +231,7 @@ def _log_spawn_metrics(self, user, options, spawn_duration_sec, spawn_exception=
                 metrics.append((metric, (date, 1)))
 
         spawn_context_key = ".".join(
-            [options[configs.lcg_rel_field], options[configs.spark_cluster_field]])
+            [options.get(configs.lcg_rel_field, "CustomEnv"), options.get(configs.spark_cluster_field, "NoSpark")])
```
etejedor, line 234: I would not touch the Spark part, I'm not sure if it can have side effects on metrics that parse it.

etejedor, line 234:
```suggestion
            [options.get(configs.lcg_rel_field, "CustomEnv"), options[configs.spark_cluster_field])
```

SwanSpawner/swanspawner/swanspawner.py
```diff
@@ -41,6 +48,22 @@ class SwanSpawner(base_class):
 
         condor_pool = 'condor-pool'
 
+        customenv_special_type = 'customenv'
+
+        eos_special_type = 'eos'
+
+        eos_pattern = Unicode(
+            default_value=r'^(?:\$CERNBOX|(?:/[^/\n]+)*/[^/\n]+)?$',
```
etejedor, line 56: How come there is no `/eos` in this pattern?

etejedor: And why $CERNBOX?

SwanSpawner/swanspawner/swanspawner.py
```diff
@@ -77,18 +100,62 @@ def __init__(self, **kwargs):
...
                 self.options_form = self._render_templated_options_form
 
+        def replace_cernbox_home(self, repo_path: str):
+            """Replace $CERNBOX_HOME with the actual path in EOS"""
+            eos_path = self.eos_path_format.format(username=self.user.name)
+            return repo_path.replace('$CERNBOX_HOME', eos_path.rstrip('/'))
+
         def options_from_form(self, formdata):
+            source_type = formdata[self.source_type][0]
+            lcg = formdata[self.lcg_rel_field][0]
```
etejedor, line 110:
```suggestion
            lcg_rel = formdata[self.lcg_rel_field][0]
```

SwanSpawner/swanspawner/swanspawner.py
```diff
@@ -77,18 +100,62 @@ def __init__(self, **kwargs):
...
+            options[self.platform_field]        = platform
             options[self.user_script_env_field] = formdata[self.user_script_env_field][0]
             options[self.spark_cluster_field]   = formdata[self.spark_cluster_field][0] if self.spark_cluster_field in formdata.keys() else 'none'
             options[self.condor_pool]           = formdata[self.condor_pool][0]
             options[self.user_n_cores]          = int(formdata[self.user_n_cores][0])
             options[self.user_memory]           = formdata[self.user_memory][0] + 'G'
 
             self.offload = options[self.spark_cluster_field] != 'none'
-
+
```
etejedor, line 158: Extra spaces.

SwanHub/swanhub/spawn_handler.py
RodrigoSobral2000: This was made for customenvs, due to the fact that lcg_rel_field would be None.
In that case, what do you think would be the best approach? Do it just if it's a lcg session?

RodrigoSobral2000: ok, nevermind, i just read the comment below

SwanSpawner/swanspawner/swanspawner.py
RodrigoSobral2000: You're right, I forgot to change this regex as you had suggested. My bad

SwanSpawner/swanspawner/swanspawner.py
```diff
@@ -77,18 +100,62 @@ def __init__(self, **kwargs):
...
+            """Replace $CERNBOX_HOME with the actual path in EOS"""
+            eos_path = self.eos_path_format.format(username=self.user.name)
+            return repo_path.replace('$CERNBOX_HOME', eos_path.rstrip('/'))
+
         def options_from_form(self, formdata):
+            source_type = formdata[self.source_type][0]
+            lcg = formdata[self.lcg_rel_field][0]
+            platform = formdata[self.platform_field][0]
+            builder_data = formdata[self.builder][0].lower().split('-')
+            builder, builder_version = (builder_data if len(builder_data) == 2 else '', '')
```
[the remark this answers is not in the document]
RodrigoSobral2000: All of them are already using that format. This was just an initial precaution, when I was not comfortable with the arguments' values returned by the form

SwanSpawner/swanspawner/swanspawner.py
```diff
@@ -77,10 +100,54 @@ def __init__(self, **kwargs):
...
                 self.options_form = self._render_templated_options_form
 
+        def replace_cernbox_home(self, repo_path: str):
+            """Replace $CERNBOX_HOME with the actual path in EOS"""
+            eos_path = self.eos_path_format.format(username=self.user.name)
+            return repo_path.replace('$CERNBOX_HOME', eos_path.rstrip('/'))
+
         def options_from_form(self, formdata):
+            source_type = formdata[self.source_type][0]
+            lcg_rel = formdata[self.lcg_rel_field][0]
```
etejedor, line 110: If custom envs is selected, what do we get here? The default LCG?

SwanSpawner/swanspawner/swanspawner.py
```diff
@@ -41,6 +48,22 @@ class SwanSpawner(base_class):
 
         condor_pool = 'condor-pool'
 
+        customenv_special_type = 'customenv'
+
+        eos_special_type = 'eos'
+
+        eos_pattern = Unicode(
+            default_value=r'^(?:\$CERNBOX_HOME|/eos(?:/[^/\n]+)*/?)?$',
```
etejedor, line 56: What is this matching exactly? Can't it be simplified?

Why do we need the first`?`?

Why do we match zero or more occurrences of `(?:\$CERNBOX_HOME|/eos(?:/[^/\n]+)*/?)`, shouldn't it be just one?

I would see if you can simplify it to one occurrence of either $CERNBOX_HOME or /eos followed by one or more patterns of "slash + folder name", isn't that enough?

SwanSpawner/swanspawner/swanspawner.py
```diff
@@ -41,6 +48,22 @@ class SwanSpawner(base_class):
...
+            config=True,
+            help='Regular expression pattern for the repository provided by a EOS folder.'
+        )
+
+        git_pattern = Unicode(
+            default_value=r'https?://(?:github\.com|gitlab\.cern\.ch)/([^/\s]+)/([^/\s]+)/?',
```
etejedor, line 62: The same patter after /eos above could be applied here after either github or gitlab, no?

SwanSpawner/swanspawner/swanspawner.py
```diff
@@ -77,10 +100,54 @@ def __init__(self, **kwargs):
...
+                    if not git_match:
+                        raise ValueError(f"Invalid Git repository: {repository}")
+
+                    repository = git_match.group(0)
+                    project_folder = git_match.group(2)
+
             options = {}
-            options[self.lcg_rel_field]         = formdata[self.lcg_rel_field][0]
-            options[self.platform_field]        = formdata[self.platform_field][0]
+            options[self.source_type]           = source_type
```
etejedor, line 143: `options[self.source_type]` should always be set, because it's telling us if we are doing LCG or CustomEnv. But shouldn't other options below only be set if it's LCG and others only if it's CustomEnv? self.builder, self.lcg_rel_field, etc.

SwanSpawner/swanspawner/swanspawner.py
```diff
@@ -106,13 +173,10 @@ def get_env(self):
 
             #FIXME remove userrid and username and just use jovyan 
             #FIXME clean JPY env variables
-            if self.lcg_rel_field in self.user_options:
+            if self.lcg_rel_field in self.user_options or self.source_type == self.customenv_special_type:
```
etejedor, line 176: Isn't this condition always true?

SwanSpawner/swanspawner/swanspawner.py
```diff
@@ -136,6 +200,13 @@ def get_env(self):
             if self.user_options[self.condor_pool] != 'none':
                 env['CERN_HTCONDOR'] = 'true'
 
+            # Enable configuration for LCG and custom environments
+            if self.user_options[self.source_type] != self.customenv_special_type:
```
etejedor, line 204: Why do we evaluate the condition negatively here? Shouldn't we have `self.customenv_special_type` and `self.lcg_special_type`? And therefore, the condition would be more readable (if the user selected lcg, I specify lcg -related vars).

SwanSpawner/swanspawner/swanspawner.py
```diff
@@ -136,6 +200,13 @@ def get_env(self):
...
+                env['ROOT_LCG_VIEW_NAME']     = self.user_options[self.lcg_rel_field]
+                env['ROOT_LCG_VIEW_PLATFORM'] = self.user_options[self.platform_field]
+                env['USER_ENV_SCRIPT']        = self.user_options[self.user_script_env_field]
+                env['ROOT_LCG_VIEW_PATH']     = self.lcg_view_path
```
etejedor, line 208: Perhaps move this one up? To be together with the LCGs.

SwanSpawner/swanspawner/swanspawner.py
```diff
@@ -136,6 +200,13 @@ def get_env(self):
             if self.user_options[self.condor_pool] != 'none':
                 env['CERN_HTCONDOR'] = 'true'
 
+            # Enable configuration for LCG and custom environments
```
etejedor, line 203:
```suggestion
            # Set LCG-related variables
```

SwanSpawner/swanspawner/templates/options_form_template.html
```diff
@@ -4,26 +4,24 @@
...
-        var lcgData;
-
-        // Data is taken from from lcgFormConfig global variable
-        var lcgFormOptions = lcgFormConfig['options'];
-        for (key in lcgFormOptions) {
-            lcgData = lcgFormOptions[key];
-            if (lcgData.type === "selection" && lcgData.lcg.value === lcgValue) {
-                return lcgData;
+    function findFormData(form, formElement, value) {
+        // Data is taken from from formConfig global variable
```
etejedor, line 15: Can you add a clearer comment on what the function does?

etejedor: And perhaps add it as doc before the function, and eliminate the comment here.

etejedor, line 15:
```suggestion
        // Data is taken from formConfig global variable
```

SwanSpawner/swanspawner/templates/options_form_template.html
```diff
@@ -41,139 +39,154 @@
...
-                if (lcgFormEntry.type === "label") {
-                    selectLCGOption.disabled="disabled";
-                    selectLCGOption.style="border-bottom:1px solid rgba(153,153,153,.3); margin:-10px 0 4px 0";
-                    selectLCGOption.value=lcgFormEntry.label.value;
-                    selectLCGOption.text=lcgFormEntry.label.text;
-                } else {
-                    if (!firstLCGSelected) {
-                        firstLCGSelected = true;
-                        selectLCGOption.selected = true;
+            for (var source in sourceConfig) {
```
etejedor, line 75: We need a comment before this loop to explain what the loop does.

SwanSpawner/swanspawner/templates/options_form_template.html
```diff
@@ -41,139 +39,154 @@
...
-            selectCondor.add(selectCondorOption);
-        }
+    /**
+     * Modifies the displayed form according to the selected type of configuration (LCG or Custom Environments)
+     */
+    function toggle_form() {
+        var lcgOption = document.getElementById('lcgOption');
+
+        var customenv_config = document.getElementById('customenv_config');
+        var external_config = document.getElementById('external_config');
```
etejedor, line 175: External is external computing resources no? Perhaps `external_res_config` better?

SwanSpawner/swanspawner/swanspawner.py
RodrigoSobral2000: the initial ?(...) is a non-capturing group that groups the contained expressions without creating backreferences, its something i used to do for good regex practices, but in this case i would say is redundant

CERNBOX_HOME points to /eos/user/u/username
and you mention this was suppose to accept anything starting with /eos /...(remaining can be anything)
I just noticed i made a mistake fixing the regex, it was supposed to be something like this in fact:
`^(?:\$CERNBOX_HOME(?:/[^/\n]+)*/?|/eos(?:/[^/\n]+)*/?)$`
wyt?

SwanSpawner/swanspawner/swanspawner.py
RodrigoSobral2000: I just did it cause I tried to avoid creating one more variable just for this need, just got the idea that we would want to declare the less amount of things possible

SwanSpawner/swanspawner/swanspawner.py
RodrigoSobral2000: I thought about that tbh, but I rather to leave them empty. This way I don't have to always check `if builder in options` for example. Its makes the code much simpler in the swan_handler module

SwanSpawner/swanspawner/swanspawner.py
RodrigoSobral2000: These /\s are used to avoid whitespaces. In the /eos path we want to avoid new lines (/\n)

SwanSpawner/swanspawner/swanspawner.py
RodrigoSobral2000: Nope, we get an empty string

RodrigoSobral2000: Since we don't want any lcg at all

SwanSpawner/swanspawner/swanspawner.py
etejedor: Why the `\n` in the expression?

etejedor: Ah, you match everything except for newlines. Is that a good pattern for matching folder names? Maybe some googling about would help.

SwanSpawner/swanspawner/swanspawner.py
etejedor: Same comment about repo folder names here.

SwanSpawner/swanspawner/swanspawner.py
etejedor: Why do you need to do `if builder in options` if you have `source_type`? Anything that depends on whether you are doing LCG or CustomEnv can use that variable as discriminant.

SwanSpawner/swanspawner/swanspawner.py
etejedor: It's true it's best to slim down what is not needed. Here I would argue it is needed in favor of readability of your code. Also, if tomorrow we add option 3 (not lcg, not customenv, something else), this becomes erroneous.

SwanSpawner/swanspawner/swanspawner.py
RodrigoSobral2000: As we're currently using the .get() method i believe you're right. Otherwise we might have to face some errors cause the logging function was accessing non existing keys (not the case anymore, indeed).

SwanHub/swanhub/spawn_handler.py
```diff
@@ -215,7 +231,7 @@ def _log_spawn_metrics(self, user, options, spawn_duration_sec, spawn_exception=
                 metrics.append((metric, (date, 1)))
 
         spawn_context_key = ".".join(
-            [options[configs.lcg_rel_field], options[configs.spark_cluster_field]])
+            [options.get(configs.lcg_rel_field, "CustomEnv"), options.get(configs.spark_cluster_field, "")])
```
etejedor, line 234: Why not leaving the spark part exactly as it was before?

etejedor: And in case we need to set a default value for it, it should be "None", since that is what the parsed metrics expect

SwanSpawner/swanspawner/swanspawner.py
```diff
@@ -41,6 +49,24 @@ class SwanSpawner(base_class):
...
         condor_pool = 'condor-pool'
 
+        customenv_special_type = 'customenv'
+
+        lcg_special_type = 'lcg'
+
+        eos_special_type = 'eos'
+
+        eos_pattern = Unicode(
+            default_value=r'^(?:\$CERNBOX_HOME|/eos)(?:/[^<>\|\\:()&;,/\n]+)*/?$',
```
etejedor, line 59: It has to be /eos/user now as per the discussion we had this morning during the meeting.

Also can you explain the regular expression?

SwanSpawner/swanspawner/swanspawner.py
```diff
@@ -41,6 +49,24 @@ class SwanSpawner(base_class):
...
+            config=True,
+            help='Regular expression pattern for the repository provided by a EOS folder.'
+        )
+
+        git_pattern = Unicode(
+            default_value=r'https?://(?:github\.com|gitlab\.cern\.ch)/([a-zA-Z0-9_-]+)/([a-zA-Z0-9_-]+)/?',
```
etejedor, line 65: Why the two final groups? User/org name plus repo name?
Then it has to be the first one without any modifier and the second one with a plus, no?

SwanSpawner/swanspawner/swanspawner.py
```diff
@@ -121,6 +190,13 @@ def get_env(self):
                     EOS_PATH_FORMAT        = self.eos_path_format,
                     SERVER_HOSTNAME        = os.uname().nodename
                 ))
+
+                # Set LCG-related variables
+                if self.user_options[self.source_type] == self.lcg_special_type:
+                    env['ROOT_LCG_VIEW_NAME']     = self.user_options[self.lcg_rel_field]
+                    env['ROOT_LCG_VIEW_PLATFORM'] = self.user_options[self.platform_field]
+                    env['USER_ENV_SCRIPT']        = self.user_options[self.user_script_env_field]
+                    env['ROOT_LCG_VIEW_PATH']     = self.lcg_view_path
```
etejedor, line 199: The else below is not correct,right? That doesn't need to be done when using customenvs. @diocas when do these variables need to be set?

etejedor (changes requested):

SwanHub/swanhub/spawn_handler.py
RodrigoSobral2000: With the implementation suggested [here](https://github.com/swan-cern/jupyterhub-extensions/pull/94#discussion_r1624549104) I noticed there was some code depending on certain lcg options, causing errors like `'spark' field not found in options`. Sometimes these errors were explicit, but not always. So, I started adopting the get() method to avoid the troubles and get the code more cautions itself. I then gave up doing and just assumed the options needed to be set no matter what, I knew I was touching code that wasn't suppose to. That's why I still set the lcg options even in a customenv configuration.

SwanSpawner/swanspawner/swanspawner.py
RodrigoSobral2000: That's what it is being done, if you look at the condition, these variables are only set when `source_type == lcg_special_type`

SwanSpawner/swanspawner/swanspawner.py
RodrigoSobral2000: Not really. Both have a plus because they gotta have at least one character. The fields/paths themselves (between slashes) are unique, mandatory and not repeatable.

SwanSpawner/swanspawner/swanspawner.py
RodrigoSobral2000: The expression must start with the env var or `/eos`
The second part says that at groups with slash `/` and least character apart from `<>\|\\:()&;,/\n` can repeat 0 or more times with an optional `/` at the end.
Now this has to be changed in a way that we can accept `CERBOX_HOME` or `/eos/user/*char*/*multiple_chars**optional_slash*`

SwanSpawner/swanspawner/swanspawner.py
RodrigoSobral2000: We must be more strict about the accepted characters in the URL due to the encoding process and how the query can be manipulated thanks to the special characters like ?, =, <, >, etc. That's why I started from a whitelist here and a blacklist (easier to control, but less freedom for naming) in the eos pattern (harder to control, but more freedom for naming)

SwanSpawner/swanspawner/swanspawner.py
etejedor: No, I mean the `else`, below. It is not correct now, it's being done for customenvs. @diocas should we just get rid of that else?

SwanSpawner/swanspawner/swanspawner.py
diocas: Contrary to some other comments I did, why don't we just validate this in the front end? If the user finds a way to bypass it, it will be properly checked in the jupyter server extension (which is the true source of truth). We might not even do any validation at all, as keeping these 2 independent components in sync requires more work than just doing it in 1 place.

SwanSpawner/swanspawner/swanspawner.py
diocas: Same as above, we might not want to validate at all.

SwanSpawner/swanspawner/swanspawner.py
```diff
@@ -77,15 +103,61 @@ def __init__(self, **kwargs):
                 # if options_form not provided, use templated options form based on configuration file
                 self.options_form = self._render_templated_options_form
 
+        def replace_cernbox_home(self, repo_path: str):
```
diocas, line 106: Why are we doing this? This should be done in the jupyter extension...

SwanSpawner/swanspawner/swanspawner.py
```diff
@@ -77,15 +103,61 @@ def __init__(self, **kwargs):
...
+            source_type = formdata[self.source_type][0]
+            lcg_rel = formdata[self.lcg_rel_field][0]
+            platform = formdata[self.platform_field][0]
+            repository, repository_type = '', ''
+            project_folder = self.user.name
+            # Builders are specified in builder-builderversion format
+            builder, builder_version = formdata[self.builder][0].lower().split('-')
+
+            if source_type == self.customenv_special_type:
+                lcg_rel, platform = '', ''
```
diocas, line 121: Instead of setting this as empty strings, you should just be adding the valid options to the `options` dict defined below.

SwanSpawner/swanspawner/swanspawner.py
```diff
@@ -77,15 +103,61 @@ def __init__(self, **kwargs):
...
+                repository = formdata[self.repository][0]
+                repository_type = formdata[self.repository_type][0]
+
+                if not repository:
+                    raise ValueError("Repository not provided")
+
+                # Do not allow the session to spawn if the repository is not valid
+                if repository_type == self.eos_special_type:
```
diocas, line 129: Just pass all the input of the user through to the place where this should be used (server extension)

SwanSpawner/swanspawner/swanspawner.py
diocas: Where are those "gets" that you mention? I really think we should never include options that are not set.

SwanSpawner/swanspawner/swanspawner.py
diocas: That else was for calls to the API instead of the form (e.g. from Binder). We don't use it, so we might consider removing it.

SwanSpawner/swanspawner/swanspawner.py
RodrigoSobral2000: You're right, I actually also thought about that as well. In matter of fact, that would be the best approach, since every verification of data related to the extension shall be coded in the extension itself, and not in the spawner, as it currently is.
However, if we don't do it in the spawner, a simple data checking would cause an error that would force the user to fill the form again and restart another session (that takes some time to create).

Making it in the spawner avoids all this frustration and the user can get the error immediately.

SwanSpawner/swanspawner/swanspawner.py
RodrigoSobral2000: In that case, could we get rid of the `if self.lcg_rel_field in self.user_options` since it will always be there?

SwanSpawner/swanspawner/swanspawner.py
RodrigoSobral2000: SwanSpawner, SwanHub and some other extensions are expecting the lcg, spark-cluster, condor, etc fields to be there. If i simply remove them from the equation because it is a customenv source, a lot of spread pieces start breaking, like the logging functions, as I mentioned. I lost some hours trying the patch them one by one until I came the conclusion I was touching code maybe I wouldn't be suppose to for unknowing its purpose.
That's I kept it simple and let them be. I came to the conclusion the relation effort-benefit wouldn't worth it cause I had other priorities.
But I might be mistaken, you tell me.

SwanHub/swanhub/spawn_handler.py
```diff
@@ -155,12 +156,27 @@ def _finish_spawn(f):
...
+                "env": configs.env_name.format(project_folder=options.get(configs.project_folder)), # {reponame}_env or {lastfolder}_env
+                "repo": options.get(configs.repository),
+            }
+            if options.get(configs.builder) == configs.accpy_special_type:
+                query_params[options.get(configs.builder)] = options.get(configs.builder_version)
+
+            # Execution SwanCustomEnvs extension with the corresponding query arguments
+            next_url = self.get_next_url(
+                user,
+                default=url_concat(url_path_join("user", user.escaped_name, "customenvs", server_name), query_params),
```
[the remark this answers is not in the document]
RodrigoSobral2000: It is not. The way the URL is passed for LCG, using base_url, takes the user to /hub/, and that's not where the extension is. that's why I forward the user to `user/rhenriqu/customenvs`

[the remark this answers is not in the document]
RodrigoSobral2000: Here's something we should discuss in the meeting

SwanSpawner/swanspawner/swanspawner.py
```diff
@@ -41,6 +47,24 @@ class SwanSpawner(base_class):
 
         condor_pool = 'condor-pool'
 
+        customenv_special_type = 'customenv'
+
+        lcg_special_type = 'lcg'
+
+        eos_special_type = 'eos'
+
+        eos_pattern = Unicode(
```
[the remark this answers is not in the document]
etejedor: In the case of the user submitting via query URL args, there is no verification on the web form. In that scenario, do we need to verify these arguments anyway before they are passed to the extension? In other words, can they cause damage for example when being added as query arguments of the extension?

SwanSpawner/swanspawner/swanspawner.py
```diff
@@ -77,15 +101,59 @@ def __init__(self, **kwargs):
...
             options = {}
-            options[self.lcg_rel_field]         = formdata[self.lcg_rel_field][0]
-            options[self.platform_field]        = formdata[self.platform_field][0]
+            options[self.source_type]           = source_type
+            options[self.user_n_cores]          = int(formdata[self.user_n_cores][0])
+            options[self.user_memory]           = formdata[self.user_memory][0] + 'G'
+            options[self.lcg_rel_field]         = lcg_rel
+            options[self.platform_field]        = platform
             options[self.user_script_env_field] = formdata[self.user_script_env_field][0]
             options[self.spark_cluster_field]   = formdata[self.spark_cluster_field][0] if self.spark_cluster_field in formdata.keys() else 'none'
```
etejedor, line 150: Check if we need this extra if else.

SwanSpawner/swanspawner/swanspawner.py
```diff
@@ -77,15 +101,59 @@ def __init__(self, **kwargs):
...
                 self.options_form = self._render_templated_options_form
 
+        def replace_cernbox_home(self, repo_path: str):
+            """Replace $CERNBOX_HOME with the actual path in EOS"""
+            eos_path = self.eos_path_format.format(username=self.user.name)
+            return repo_path.replace('$CERNBOX_HOME', eos_path.rstrip('/'))
+
         def options_from_form(self, formdata):
+            source_type = formdata[self.source_type][0]
+            lcg_rel = formdata[self.lcg_rel_field][0]
```
etejedor, line 111: Is this `none` if we go through the CustomEnv path?

SwanSpawner/swanspawner/swanspawner.py
```diff
@@ -77,15 +101,59 @@ def __init__(self, **kwargs):
...
+
+                    repository = git_match.group(0)
+
             options = {}
-            options[self.lcg_rel_field]         = formdata[self.lcg_rel_field][0]
-            options[self.platform_field]        = formdata[self.platform_field][0]
+            options[self.source_type]           = source_type
+            options[self.user_n_cores]          = int(formdata[self.user_n_cores][0])
+            options[self.user_memory]           = formdata[self.user_memory][0] + 'G'
+            options[self.lcg_rel_field]         = lcg_rel
```
etejedor, line 147: Check what happens if you don't set this option when selecting a CustomEnv.

SwanSpawner/swanspawner/swanspawner.py
```diff
@@ -77,15 +101,59 @@ def __init__(self, **kwargs):
...
+            options[self.platform_field]        = platform
```
etejedor, line 148: Same here, check what happens with this in CustomEnv.

SwanSpawner/swanspawner/swanspawner.py
```diff
@@ -106,31 +174,23 @@ def get_env(self):
...
-                env.update(dict(
-                    USER                   = "jovyan",
-                    HOME                   = "/home/jovyan",
-                    NB_USER                = 'jovyan',
-                    USER_ID                = 1000,
-                    NB_UID                 = 1000,
-                    SERVER_HOSTNAME        = os.uname().nodename,
-                ))
+            env.update(dict(
+                SOURCE_TYPE            = self.user_options[self.source_type],
```
etejedor, line 178:
```suggestion
                SOFTWARE_SOURCE            = self.user_options[self.software_source],
```

etejedor (changes requested):

SwanSpawner/swanspawner/swanspawner.py
etejedor: The information about LCG releases is obtained here (the defaults) even if the user selects a CustomEnv.

SwanSpawner/swanspawner/swanspawner.py
```diff
@@ -77,15 +101,57 @@ def __init__(self, **kwargs):
...
+
             options = {}
-            options[self.lcg_rel_field]         = formdata[self.lcg_rel_field][0]
-            options[self.platform_field]        = formdata[self.platform_field][0]
+            options[self.software_source]       = software_source
+            options[self.user_n_cores]          = int(formdata[self.user_n_cores][0])
+            options[self.user_memory]           = formdata[self.user_memory][0] + 'G'
+            options[self.lcg_rel_field]         = lcg_rel
+            options[self.platform_field]        = platform
             options[self.user_script_env_field] = formdata[self.user_script_env_field][0]
```
etejedor, line 147: And this.

PR #101, opened 2024-07-03
SwanSpawner: Add metric to monitor the usage of each interface

PMax5:
Add log metric to check the number of users that start a session with the classic and jupyterlab interfaces.

### DIFF
--- a/SwanSpawner/swanspawner/swanspawner.py
+++ b/SwanSpawner/swanspawner/swanspawner.py
@@ -144,6 +144,17 @@ def define_SwanSpawner_from(base_class):
                     SWAN_USE_JUPYTERLAB = 'true'
                 ))
 
+                user_interface = 'lab'
+            else:
+                user_interface = 'classic'
+
+            self.log_metric(
+                self.user.name,
+                self.this_host,
+                'ui',
+                user_interface
+            )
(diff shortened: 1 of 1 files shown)


SwanSpawner/swanspawner/swanspawner.py
```diff
@@ -144,6 +144,17 @@ def get_env(self):
                     SWAN_USE_JUPYTERLAB = 'true'
                 ))
 
+                user_interface = "lab"
```
etejedor, line 147: I'ts a minor thing, but can you keep the style of the single quotes for string literals?

SwanSpawner/swanspawner/swanspawner.py
```diff
@@ -144,6 +144,17 @@ def get_env(self):
...
                 ))
 
+                user_interface = 'lab'
+            else:
+                user_interface = 'classic'
+
+            self.log_metric(
+                self.user.name,
+                self.this_host,
+                "ui",
```
etejedor, line 154: Also here please

PMax5: Fixed!

etejedor (approved):

PR #104, opened 2024-07-08
New feature for custom environments

RodrigoSobral2000:
Result:
Spawner can now launch custom environments through its web interface.

Developments:
Thanks to changes on SwanSpawner and SwanHub, its now possible to avoid having to install all LCG packages and provide a specific amount of packages through cernbox/eos path or git repo containing a requirements.txt file.

This new custom virtual environment can then be launched by a generic python venv or a acc-py stack.

The user gets redirected to /customenv where the process details are shown up in real time, so the users is able to keep up with its going on under the hood. In the end, a completely empty custom environment (just with the provided packages) gets ready for the user to work on.

### DIFF
--- a/SwanHub/swanhub/spawn_handler.py
+++ b/SwanHub/swanhub/spawn_handler.py
@@ -107,6 +107,7 @@ class SpawnHandler(JHSpawnHandler):
 
         start_time_spawn = time.time()
 
+        options = {}
         try:
             options = await maybe_future(spawner.run_options_from_form(form_options))
             await self.spawn_single_user(user, server_name=server_name, options=options)
@@ -136,7 +137,6 @@ class SpawnHandler(JHSpawnHandler):
                     user, options, time.time() - start_time_spawn)
 
         except Exception as e:
-
             self._log_spawn_metrics(
                 user, options, time.time() - start_time_spawn, e)
 
@@ -155,12 +155,21 @@ class SpawnHandler(JHSpawnHandler):
 
         if current_user is user:
             self.set_login_cookie(user)
-        next_url = self.get_next_url(
-            user,
-            default=url_path_join(
-                self.hub.base_url, "spawn-pending", user.escaped_name, server_name
-            ),
-        )
+
+        if options.get(configs.software_source) == configs.customenv_special_type:
+            # Add the query parameters to the URL
+            query_params = {
+                "repo": options.get(configs.repository),
+                "repo_type": options.get(configs.repo_type),
+            }
+            if options.get(configs.builder) == configs.accpy_special_type:
+                query_params[options.get(configs.builder)] = options.get(configs.builder_version)
+
+            # Execution SwanCustomEnvs extension with the corresponding query arguments
+            next_url = url_concat(url_path_join("user", user.escaped_name, "customenvs", server_name), query_params)
+        else: # LCG release
+            next_url = url_path_join(self.hub.base_url, "spawn-pending", user.escaped_name, server_name)
+
         self.redirect(next_url)
 
     async def _render_form_wrapper(self, for_user, message=''):
@@ -215,7 +224,7 @@ class SpawnHandler(JHSpawnHandler):
                 metrics.append((metric, (date, 1)))
 
         spawn_context_key = ".".join(
-            [options[configs.lcg_rel_field], options[configs.spark_cluster_field]])
+            [options.get(configs.lcg_rel_field, "CustomEnv"), options.get(configs.spark_cluster_field, "none")])
         if not spawn_exception:
             # Add spawn success (no exception) and duration to the log and send as metrics
             spawn_exc_class = "None"
--- a/SwanSpawner/swanspawner/swanspawner.py
+++ b/SwanSpawner/swanspawner/swanspawner.py
@@ -112,8 +112,8 @@ def define_SwanSpawner_from(base_class):
                 options[self.repository]      = formdata[self.repository][0]
                 options[self.repo_type]       = formdata[self.repo_type][0]
 
-                if not options[self.repository]:
-                    raise ValueError("No Repository specified")
+                if options[self.repository] is None or options[self.repository].strip() == '':
+                    raise ValueError('No Repository specified')
             else:
                 options[self.lcg_rel_field]         = formdata[self.lcg_rel_field][0]
                 options[self.platform_field]        = formdata[self.platform_field][0]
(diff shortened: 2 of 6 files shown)


SwanSpawner/swanspawner/swanspawner.py
```diff
@@ -82,18 +97,31 @@ def __init__(self, **kwargs):
...
             options[self.user_memory]               = formdata[self.user_memory][0] + 'G'
             options[self.use_jupyterlab_field]      = formdata.get(self.use_jupyterlab_field, 'unchecked')[0]
             options[self.use_local_packages_field]  = formdata.get(self.use_local_packages_field, 'unchecked')[0]
+            if options[self.software_source] == self.customenv_special_type:
+                options[self.builder]         = builder
+                options[self.builder_version] = builder_version
+                options[self.repository]      = formdata[self.repository][0]
+                options[self.repo_type]       = formdata[self.repo_type][0]
+
+                if not options[self.repository]:
```
etejedor, line 115: So here you get an empty string if the user does not specify any repository URL?

RodrigoSobral2000: if the user doesn't specify any repo (url or eos path) he gets an instant error message. the idea is to ensure we get a value to send to the extension

SwanSpawner/swanspawner/swanspawner.py
```diff
@@ -82,18 +97,31 @@ def __init__(self, **kwargs):
...
+                    raise ValueError("No Repository specified")
```
etejedor, line 116: Is this shown in red in the web form? If so, please check my suggestion below.

etejedor, line 116: `raise ValueError('')`

Note also the single quotes, please keep the style for string literals.

SwanHub/swanhub/spawn_handler.py
```diff
@@ -155,12 +155,21 @@ def _finish_spawn(f):
...
+        if options.get(configs.software_source) == configs.customenv_special_type:
+            # Add the query parameters to the URL
+            query_params = {
+                "repo": options.get(configs.repository),
+                "repo_type": options.get(configs.repo_type),
+            }
+            if options.get(configs.builder) == configs.accpy_special_type:
+                query_params[options.get(configs.builder)] = options.get(configs.builder_version)
+
+            # Execution SwanCustomEnvs extension with the corresponding query arguments
```
etejedor, line 168: @diocas could you please do a final check of this?

SwanSpawner/swanspawner/swanspawner.py
RodrigoSobral2000: If I remember correctly, you get a None value.

etejedor: Then the right way to write that condition is:
`if options[self.repository] is None:`

Since Python variables do not have a type, it is better to check conditions against exactly the type you expect, to prevent unforeseen behavior.

SwanSpawner/swanspawner/templates/options_form_template.html
```diff
@@ -224,7 +207,61 @@
...
+        const width = tempElement.offsetWidth;
+        document.body.removeChild(tempElement);
+
+        selectElement.style.width = `${width + 50}px`;
+
+        const repositoryOption = document.getElementById('repositoryOption');
+        if (selectElement.value === 'eos') {
+            repositoryOption.placeholder = 'e.g. $CERNBOX_HOME/MyFolder';
+            // Regular expression pattern for the repository provided by a EOS folder.
+            repositoryOption.pattern = '^(\\\$CERNBOX_HOME(\\/[^<>\|\\\\:()&;,\n]+)*\\/?|\\/eos\\/user\\/[a-z](\\/[^<>\|\\\\:()&;,\n]+)+\\/?)$';
```
etejedor, line 232: What happens if the URL does not fulfill the pattern? Is it still possible to submit the form?
@diocas please have a look.

SwanSpawner/swanspawner/swanspawner.py
```diff
@@ -112,8 +112,8 @@ def options_from_form(self, formdata):
                 options[self.repository]      = formdata[self.repository][0]
                 options[self.repo_type]       = formdata[self.repo_type][0]
 
-                if not options[self.repository]:
-                    raise ValueError("No Repository specified")
+                if options[self.repository] is None or options[self.repository].strip() == '':
```
etejedor, line 115: But is it `None` or a string?

RodrigoSobral2000: It is None, but I thought that a user could write multiple spaces. In that case, it gets a red error message as well

SwanSpawner/swanspawner/swanspawner.py
```diff
@@ -82,18 +97,31 @@ def __init__(self, **kwargs):
...
             options[self.use_jupyterlab_field]      = formdata.get(self.use_jupyterlab_field, 'unchecked')[0]
             options[self.use_local_packages_field]  = formdata.get(self.use_local_packages_field, 'unchecked')[0]
+            if options[self.software_source] == self.customenv_special_type:
+                options[self.builder]         = builder
+                options[self.builder_version] = builder_version
+                options[self.repository]      = formdata[self.repository][0]
+                options[self.repo_type]       = formdata[self.repo_type][0]
+
+                if options[self.repository] is None or options[self.repository].strip() == '':
+                    raise ValueError('No Repository specified')
```
etejedor, line 116: This did not change according to my comment.

SwanSpawner/swanspawner/templates/options_form_template.html
RodrigoSobral2000: Yes, that is just a visual guidance for the user to know what it is expected from the form

SwanSpawner/swanspawner/swanspawner.py
etejedor: But then you are checking another condition, which is different from the message you throw (you throw repository not specified). Please just keep the None check, the other check will be done in the backend script as agreed (no further checks here in the spawner).

SwanSpawner/swanspawner/swanspawner.py
RodrigoSobral2000: Did you really mean `raise ValueError('')` with an empty message?

etejedor: No, it was not an empty message. It had `raise ValueError('Cannot create custom software environment: no repository specified')`

SwanSpawner/swanspawner/templates/options_form_template.html
etejedor: Ok, let's go with this for now. If @diocas knows how to block the submit of the form (or make it fail immediately?), he can tell us how to do it.

etejedor (approved):

PR #105, opened 2024-07-10
Fix on if statement regarding empty repository error

RodrigoSobral2000:
When input fields are empty, the value that is sent to backend is an empty string and not a None value. This PR fix that verification


SwanSpawner/swanspawner/swanspawner.py
```diff
@@ -112,7 +112,7 @@ def options_from_form(self, formdata):
                 options[self.repository]      = formdata[self.repository][0]
                 options[self.repo_type]       = formdata[self.repo_type][0]
 
-                if options[self.repository] is None:
+                if options[self.repository] == '':
```
etejedor, line 115: `if not options[self.repository]` --> would be better to use the boolean value of strings if we know this is going to be a string.

PR #107, opened 2024-07-10
New environment variable defined by spawner

RodrigoSobral2000:
`SOFTWARE_SOURCE` is used for the script on `swan` image layer to know if it has to configure lcg or not

### DIFF
--- a/SwanSpawner/swanspawner/swanspawner.py
+++ b/SwanSpawner/swanspawner/swanspawner.py
@@ -141,6 +141,7 @@ def define_SwanSpawner_from(base_class):
             #FIXME remove userrid and username and just use jovyan 
             #FIXME clean JPY env variables
             env.update(dict(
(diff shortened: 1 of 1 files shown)


etejedor:
Some comments:
- For every PR you open that is your responsibility (your changes), assign it to yourself and assign one or more reviewers.
- This PR has two commits. The first commit makes some change, and the second commit applies a change to that same line to fix a bug in the previous commit. A PR cannot contain commits that fix bugs introduced by commits of the same PR. If you found a problem with the first commit, you need to fix it a second commit, then do an interactive rebase and merge those two commits into one (see fixup option of git rebase -i).
- Please do not ask reviewers to review PRs that have as a commit message things like "HOTFIX" in this case. This is something you can easily avoid before you ask anyone to see your commits, and it reduces the review effort of reviewers. Please always check, before asking anyone to review, that the commits of your PR have the changes you want and the right messages.

diocas (approved):

PR #109, opened 2024-07-16
swanhub, swanspawner: New feature for notebook arguments

RodrigoSobral2000:
Spawner can now use GET for getting the same arguments it gets through POST requests. With this, an URL that includes a `notebook` argument can open it after launching a session with the remaining specified parameters. If it is a customenv session, it just forwards the `notebook` argument to the SwanCustomEnvironments extension and this will take care of the rest.

### DIFF
--- a/SwanHub/swanhub/spawn_handler.py
+++ b/SwanHub/swanhub/spawn_handler.py
@@ -171,7 +171,7 @@ class SpawnHandler(JHSpawnHandler):
             query_params = {
                 "repo": options.get(configs.repository),
                 "repo_type": options.get(configs.repo_type),
-                "notebook": options.get(configs.notebook, ''),
+                "file": options.get(configs.file, ''),
             }
             if options.get(configs.builder) == configs.accpy_special_type:
                 query_params[options.get(configs.builder)] = options.get(configs.builder_version)
@@ -180,8 +180,8 @@ class SpawnHandler(JHSpawnHandler):
             next_url = url_concat(url_path_join("user", user.escaped_name, "customenvs", server_name), query_params)
         else: # LCG release
             next_url = url_path_join(self.hub.base_url, "spawn-pending", user.escaped_name, server_name)
-            if options.get(configs.notebook) and options[configs.use_jupyterlab_field] == 'checked':
-                next_url = url_path_join("user", user.escaped_name, "lab", "tree", *options[configs.notebook].split('/'))
+            if options.get(configs.file) and options[configs.use_jupyterlab_field] == 'checked':
+                next_url = url_path_join("user", user.escaped_name, "lab", "tree", *options[configs.file].split('/'))
 
         self.redirect(next_url)
 
--- a/SwanHub/swanhub/spawn_handler.py
+++ b/SwanHub/swanhub/spawn_handler.py
@@ -54,6 +54,15 @@ class SpawnHandler(JHSpawnHandler):
             self.finish(form)
             return
 
+        # By default, the super().get() starts spawning the session if any
+        # query arguments are specified. So, we need to explicitly check for
+        # the software_source query argument to re-render the form and fulfill
+        # it with the remaining query arguments.
+        if configs.software_source in self.request.query_arguments:
+            form = await self._render_form_wrapper(user)
+            self.finish(form)
+            return
+
         try:
             await super().get(for_user, server_name)
         except web.HTTPError as e:
@@ -156,11 +165,13 @@ class SpawnHandler(JHSpawnHandler):
         if current_user is user:
             self.set_login_cookie(user)
 
+
         if options.get(configs.software_source) == configs.customenv_special_type:
             # Add the query parameters to the URL
             query_params = {
                 "repo": options.get(configs.repository),
                 "repo_type": options.get(configs.repo_type),
+                "notebook": options.get(configs.notebook, ''),
             }
             if options.get(configs.builder) == configs.accpy_special_type:
                 query_params[options.get(configs.builder)] = options.get(configs.builder_version)
@@ -169,6 +180,8 @@ class SpawnHandler(JHSpawnHandler):
             next_url = url_concat(url_path_join("user", user.escaped_name, "customenvs", server_name), query_params)
         else: # LCG release
             next_url = url_path_join(self.hub.base_url, "spawn-pending", user.escaped_name, server_name)
+            if options.get(configs.notebook) and options[configs.use_jupyterlab_field] == 'checked':
+                next_url = url_path_join("user", user.escaped_name, "lab", "tree", *options[configs.notebook].split('/'))
 
         self.redirect(next_url)
 
--- a/SwanHub/swanhub/spawn_handler.py
+++ b/SwanHub/swanhub/spawn_handler.py
@@ -56,7 +56,7 @@ class SpawnHandler(JHSpawnHandler):
 
         # By default, the super().get() starts spawning the session if any
         # query arguments are specified. So, we need to explicitly check for
-        # the software_source query argument to re-render the form and fulfill
+        # the software_source query argument to re-render the form and fill
         # it with the remaining query arguments.
         if configs.software_source in self.request.query_arguments:
             form = await self._render_form_wrapper(user)
--- a/SwanSpawner/swanspawner/swanspawner.py
+++ b/SwanSpawner/swanspawner/swanspawner.py
@@ -54,6 +54,8 @@ def define_SwanSpawner_from(base_class):
 
         condor_pool = 'condor-pool'
 
+        notebook = 'notebook'
+
         customenv_special_type = 'customenv'
 
         lcg_special_type = 'lcg'
@@ -106,6 +108,8 @@ def define_SwanSpawner_from(base_class):
             options[self.user_memory]               = formdata[self.user_memory][0] + 'G'
             options[self.use_jupyterlab_field]      = formdata.get(self.use_jupyterlab_field, 'unchecked')[0]
             options[self.use_local_packages_field]  = formdata.get(self.use_local_packages_field, 'unchecked')[0]
+            options[self.notebook]                  = formdata.get(self.notebook, '')[0]
+
             if options[self.software_source] == self.customenv_special_type:
                 options[self.builder]         = builder
                 options[self.builder_version] = builder_version
--- a/SwanSpawner/swanspawner/swanspawner.py
+++ b/SwanSpawner/swanspawner/swanspawner.py
@@ -54,7 +54,7 @@ def define_SwanSpawner_from(base_class):
 
         condor_pool = 'condor-pool'
 
-        notebook = 'notebook'
+        file = 'file'
 
         customenv_special_type = 'customenv'
 
@@ -108,7 +108,7 @@ def define_SwanSpawner_from(base_class):
             options[self.user_memory]               = formdata[self.user_memory][0] + 'G'
             options[self.use_jupyterlab_field]      = formdata.get(self.use_jupyterlab_field, 'unchecked')[0]
             options[self.use_local_packages_field]  = formdata.get(self.use_local_packages_field, 'unchecked')[0]
-            options[self.notebook]                  = formdata.get(self.notebook, '')[0]
+            options[self.file]                      = formdata.get(self.file, '')[0]
 
             if options[self.software_source] == self.customenv_special_type:
                 options[self.builder]         = builder
--- a/SwanSpawner/swanspawner/templates/options_form_template.html
+++ b/SwanSpawner/swanspawner/templates/options_form_template.html
@@ -40,7 +40,7 @@
     /**
      * Get the value of the first selectable option in the given select element
      */
-    function select_default_option(id) {
+    function get_default_option(id) {
         return document.getElementById(id).selectedOptions[0].value;
     }
     /**
--- a/SwanSpawner/swanspawner/templates/options_form_template.html
+++ b/SwanSpawner/swanspawner/templates/options_form_template.html
@@ -46,7 +46,7 @@
     /**
      * Set the given input with the specified value
      */
-    function select_input(id, value) {
+    function set_input(id, value) {
         document.getElementById(id).value = value;
     }
     /**
--- a/SwanSpawner/swanspawner/templates/options_form_template.html
+++ b/SwanSpawner/swanspawner/templates/options_form_template.html
@@ -8,6 +8,19 @@
 
     var formInitialized = false;
 
+    var sourceConfig = {
+        lcg: {
+            optionsElement: "lcgReleaseOptions",
+            sourceForm: "lcg_options",
+            formElement: "lcg",
+        },
+        customenv: {
+            optionsElement: "builderOptions",
+            sourceForm: "customenv_options",
+            formElement: "builder",
+        }
+    };
+
     /**
      * Get Form data dictionary for given selectedValue.
      */
@@ -25,6 +38,33 @@
     }
 
     /**
+     * Get the value of the first selectable option in the given select element
+     */
+    function select_default_option(id) {
+        return document.getElementById(id).selectedOptions[0].value;
+    }
+    /**
+     * Fulfill the given input with the specified value
+     */
+    function select_input(id, value) {
+        document.getElementById(id).value = value;
+    }
+    /**
+     * Loops throught the elements keys and selects the option with the given value
+     */
+    function select_options(elements) {
+        for (var id in elements) {
+            select_input(id, elements[id]);
+        }
+    }
+    /**
+     * (Un)check the element with the given id
+     */
+    function select_checkbox(id, checked) {
+        document.getElementById(id).checked = checked;
+    }
+
+    /**
      * Enable display of element if previously disabled, and disable if previously enabled
      */
     function toggle_visibility(id) {
@@ -38,7 +78,7 @@
     /**
      * Enable display of element if previously disabled, and disable if previously enabled
      */
-     function validate_repository_input() {
+    function validate_repository_input() {
         var repoInput = document.getElementById('repositoryOption');
         const repoPattern = new RegExp(repoInput.pattern);
 
@@ -53,27 +93,7 @@
      * Renders a form based on config file
      */
     function adjust_form() {
-        const sourceOptions = document.getElementsByName('software_source');
-        let selectedSource = null;
-        for (let i = 0; i < sourceOptions.length; i++) {
-            if (sourceOptions[i].checked) {
-                selectedSource = sourceOptions[i].value;
-                break;
-            }
-        }
-
-        var sourceConfig = {
-            lcg: {
-                optionsElement: document.getElementById("lcgReleaseOptions"),
-                sourceForm: "lcg_options",
-                formElement: "lcg",
-            },
-            customenv: {
-                optionsElement: document.getElementById("builderOptions"),
-                sourceForm: "customenv_options",
-                formElement: "builder",
-            }
-        };
+        let selectedSource = document.querySelector('input[name="software_source"]:checked').value;
 
         // if form has not been yet initialized, initialize
         if (!formInitialized) {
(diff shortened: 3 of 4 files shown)


SwanHub/swanhub/spawn_handler.py
```diff
@@ -54,6 +55,24 @@ async def get(self, for_user=None, server_name=''):
             self.finish(form)
             return
 
+        if configs.software_source in self.request.query_arguments:
+            query_options = {key: [value[0].encode("utf-8")] for key, value in parse_qs(urlparse(unquote(self.request.uri)).query).items()}
+            self.request.body_arguments = {
+                configs.software_source: query_options.get(configs.software_source, [b'']),
```
diocas, line 61: Why are defaults arrays of bytes?

SwanSpawner/swanspawner/swanspawner.py
```diff
@@ -106,6 +111,12 @@ def options_from_form(self, formdata):
             options[self.user_memory]               = formdata[self.user_memory][0] + 'G'
             options[self.use_jupyterlab_field]      = formdata.get(self.use_jupyterlab_field, 'unchecked')[0]
             options[self.use_local_packages_field]  = formdata.get(self.use_local_packages_field, 'unchecked')[0]
+
+            # Clear notebook path from CERNBOX_HOME and EOS path format
+            options[self.notebook]                  = formdata.get(self.notebook, [''])[0]
+            if options[self.notebook] and not match(self.notebook_pattern, options[self.notebook]):
+                raise ValueError('Invalid notebook path: %s' % options[self.notebook])
```
diocas, line 118: f-strings is better.

diocas (commented):
So... I have opinions...

I don't know what was discussed with @etejedor , but why are we setting this logic in the backend and not in the frontend? (here we go again, always frontend.. I know..)

My big issue with this is that we are adding logic to the spawn handler when we should be removing it and stop wrapping jupyterhub.

I suppose the reason to add it to the backend was to make the startup automatic. If it was in the frontend, users get the options automatically selected, but can still do changes before starting the session. Also, they can see what are the options that are being selected.
For example, with the current implementation, I can create a short link to SWAN, with a user script parameter pointing to a script in my home, that fetches the oauth token of the users. If someone clicks on it, this would start automatically without having time to check if it is safe.
Also, until we do backend checks of the options (that we should do), this makes it easier for users to try non-accepted options.

If we still want to make the startup automatic, I would check first if we can interact with the spawner from a service (I think we can't, but I would check it first). If we can, we could expose a new endpoint that gets the parameters and starts the session. It can even have a confirmation dialogue asking the user if they trust the options.

SwanSpawner/swanspawner/swanspawner.py
```diff
@@ -106,6 +111,12 @@ def options_from_form(self, formdata):
             options[self.user_memory]               = formdata[self.user_memory][0] + 'G'
             options[self.use_jupyterlab_field]      = formdata.get(self.use_jupyterlab_field, 'unchecked')[0]
             options[self.use_local_packages_field]  = formdata.get(self.use_local_packages_field, 'unchecked')[0]
+
+            # Clear notebook path from CERNBOX_HOME and EOS path format
+            options[self.notebook]                  = formdata.get(self.notebook, [''])[0]
```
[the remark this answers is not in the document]
RodrigoSobral2000: The `None` value will be propagated to the script and it will be read as a normal string. If empty, it gets easier for the script to check if the its value is empty.

SwanSpawner/swanspawner/templates/options_form_template.html
```diff
@@ -430,5 +511,18 @@ <h2>External computing resources</h2>
             </label>
             <select id="condorOptions" name="condor-pool"></select>
         </div>
+        <br>
+    </div>
+
+    <!-- Notebook path information text -->
```
diocas, line 517: Not a fan, as you know :) We can wait to see what @etejedor thinks.

SwanSpawner/swanspawner/templates/options_form_template.html
```diff
@@ -242,23 +269,77 @@
      */
      function toggle_form() {
         var lcgOption = document.getElementById('lcgOption');
+        var customenvOption = document.getElementById('customenvOption');
 
         var customenv_config = document.getElementById('customenv_config');
         var external_res_config = document.getElementById('external_res_config');
         var lcg_config = document.getElementById('lcg_config');
+        var notebook_config = document.getElementById('notebook_config');
```
diocas, line 277: Not sure if `notebook_config` conveys the real meaning of this var/text box. It's just the notebook url, right?

SwanSpawner/swanspawner/templates/options_form_template.html
```diff
@@ -242,23 +269,77 @@
...
 
         adjust_form();
+
+        const urlParams = new URLSearchParams(window.location.search);
+        const software_source = urlParams.get('software_source');
+        const is_software_source_valid = ["lcg", "customenv"].includes(software_source);
```
diocas, line 283: Take `sourceConfig` out of the function where is defined, and put it in the parent scope so that you can use it here as the list of valid sources. Otherwise you have 2 lists that need to be kept in sync.

SwanSpawner/swanspawner/templates/options_form_template.html
```diff
@@ -242,23 +269,77 @@
...
+        if (is_software_source_valid) {
+            const selected_cores = urlParams.get('ncores') || select_default_option("coresOptions");
+            const selected_memory = urlParams.get('memory') || select_default_option("memoryOptions");
+            const notebook = urlParams.get('notebook') || "";
+            const use_jupyterlab = (urlParams.get('use-jupyterlab') || "").toLowerCase() === 'true';
```
diocas, line 288: Not really a choice if you redirect users to lab when they provide a notebook path.

SwanSpawner/swanspawner/templates/options_form_template.html
```diff
@@ -242,23 +269,77 @@
...
         var external_res_config = document.getElementById('external_res_config');
         var lcg_config = document.getElementById('lcg_config');
+        var notebook_config = document.getElementById('notebook_config');
 
         adjust_form();
+
+        const urlParams = new URLSearchParams(window.location.search);
+        const software_source = urlParams.get('software_source');
+        const is_software_source_valid = ["lcg", "customenv"].includes(software_source);
+        if (is_software_source_valid) {
```
diocas, line 284: Should we show an error?

SwanSpawner/swanspawner/templates/options_form_template.html
```diff
@@ -242,23 +269,77 @@
...
+
         if (lcgOption.checked) {
             lcg_config.style.display = 'block';
             customenv_config.style.display = 'none';
             external_res_config.style.display = 'block';
-            adjust_spark();
+
+            if (is_software_source_valid) {
+                customenvOption.disabled = true;
+                const selected_lcg = urlParams.get('LCG-rel') || select_default_option("lcgReleaseOptions");
```
diocas, line 306: Do you really need to have so many similar lines? I would try to use a for-loop (as discussed), and you could even get the parameters directly from url params without having to define them here. I mean, a dynamic thing, so that I can add a form field called cenasOptions, and to set its value, I would just need to provide a query param called "cenas", without ever touching this code here.

SwanSpawner/swanspawner/templates/options_form_template.html
```diff
@@ -242,23 +269,77 @@
...
+        }
+
         if (lcgOption.checked) {
             lcg_config.style.display = 'block';
             customenv_config.style.display = 'none';
             external_res_config.style.display = 'block';
-            adjust_spark();
+
+            if (is_software_source_valid) {
+                customenvOption.disabled = true;
```
diocas, line 305: This is not very future-proof. If I need to create a different form section, I would have to set both as disabled. Doing the opposite, making the default disabled and enabling it on selection, would make it easier.

diocas (changes requested):

SwanSpawner/swanspawner/templates/options_form_template.html
RodrigoSobral2000: More than just that, it includes the label, `more...` details and the input with the notebook value itself

RodrigoSobral2000: I tried to follow the same pattern as the ones above

SwanHub/swanhub/spawn_handler.py
```diff
@@ -54,6 +54,15 @@ async def get(self, for_user=None, server_name=''):
             self.finish(form)
             return
 
+        # By default, the super().get() starts spawning the session if any
+        # query arguments are specified. So, we need to explicitly check for
+        # the software_source query argument to re-render the form and fulfill
```
diocas, line 59:
```suggestion
        # the software_source query argument to re-render the form and fill
```

SwanSpawner/swanspawner/templates/options_form_template.html
```diff
@@ -24,6 +37,33 @@
         throw `${formElement} Data not found`
     }
 
+    /**
+     * Get the value of the first selectable option in the given select element
+     */
+    function select_default_option(id) {
```
diocas, line 43:
```suggestion
    function get_default_option(id) {
```

SwanSpawner/swanspawner/templates/options_form_template.html
```diff
@@ -24,6 +37,33 @@
...
+        return document.getElementById(id).selectedOptions[0].value;
+    }
+    /**
+     * Fulfill the given input with the specified value
```
diocas, line 47:
```suggestion
     * Set the given input with the specified value
```

SwanSpawner/swanspawner/templates/options_form_template.html
```diff
@@ -24,6 +37,33 @@
...
+     */
+    function select_input(id, value) {
```
diocas, line 49:
```suggestion
    function set_input(id, value) {
```

SwanSpawner/swanspawner/templates/options_form_template.html
```diff
@@ -24,6 +37,33 @@
...
+        document.getElementById(id).value = value;
+    }
+    /**
+     * Loops throught the elements keys and selects the option with the given value
+     */
+    function select_options(elements) {
```
diocas, line 55:
```suggestion
    function set_options(elements) {
```

SwanSpawner/swanspawner/templates/options_form_template.html
```diff
@@ -24,6 +37,33 @@
...
+        for (var id in elements) {
+            select_input(id, elements[id]);
+        }
+    }
+    /**
+     * (Un)check the element with the given id
+     */
+    function select_checkbox(id, checked) {
```
diocas, line 63:
```suggestion
    function set_checkbox(id, checked) {
```

SwanSpawner/swanspawner/templates/options_form_template.html
```diff
@@ -246,19 +266,76 @@
...
+                const selected = urlParams.get(param) || select_default_option(elem);
+                select_input(elem, selected);
+            }
+            for (const [param, elem] of Object.entries(query_map.inputs)) {
+                const value = urlParams.get(param) || "";
+                select_input(elem, value);
+            }
+            notebook_config.style.display = 'block';
+        } else if (software_source && !is_software_source_valid) {
+            const error_message = document.createElement('span');
```
diocas, line 314: Can we make the error build a function?

SwanSpawner/swanspawner/templates/options_form_template.html
```diff
@@ -246,19 +266,76 @@
...
+            error_message.style.color = 'red';
+            error_message.innerHTML = `Invalid software source: ${software_source}`;
+            error_message.appendChild(document.createElement('br'));
```
diocas, line 317: why a br?

SwanSpawner/swanspawner/templates/options_form_template.html
```diff
@@ -246,19 +266,76 @@
...
+            if (!optionsHeader.querySelector('span')) {
+                optionsHeader.insertBefore(error_message, optionsHeader.firstChild);
+            }
+        }
+
         if (lcgOption.checked) {
             lcg_config.style.display = 'block';
             customenv_config.style.display = 'none';
             external_res_config.style.display = 'block';
-            adjust_spark();
```
diocas, line 255: no need to adjust spark?

diocas (changes requested):
Just some small suggestions and we're good to go.

SwanSpawner/swanspawner/swanspawner.py
```diff
@@ -97,23 +99,26 @@ def __init__(self, **kwargs):
...
-            # Builders are specified in builder-builderversion format
-            builder, builder_version = formdata[self.builder][0].lower().split('-')
-
             options = {}
             options[self.software_source]           = formdata[self.software_source][0]
             options[self.user_n_cores]              = int(formdata[self.user_n_cores][0])
             options[self.user_memory]               = formdata[self.user_memory][0] + 'G'
             options[self.use_jupyterlab_field]      = formdata.get(self.use_jupyterlab_field, 'unchecked')[0]
             options[self.use_local_packages_field]  = formdata.get(self.use_local_packages_field, 'unchecked')[0]
+            options[self.notebook]                  = formdata.get(self.notebook, '')[0]
```
diocas, line 108: Since we allow any file, "notebook" might be a bit misleading.

SwanSpawner/swanspawner/templates/options_form_template.html
RodrigoSobral2000: From what I understand, `adjust_spark` was renamed to `adjust_options`, and it is being called at line 330 in order to update its options according to the freshly set lcg/platform

SwanSpawner/swanspawner/templates/options_form_template.html
RodrigoSobral2000: To create a separation between the message error and the header message

SwanSpawner/swanspawner/swanspawner.py
```diff
@@ -97,23 +99,26 @@ def __init__(self, **kwargs):
...
-            # Builders are specified in builder-builderversion format
-            builder, builder_version = formdata[self.builder][0].lower().split('-')
-
             options = {}
             options[self.software_source]           = formdata[self.software_source][0]
             options[self.user_n_cores]              = int(formdata[self.user_n_cores][0])
             options[self.user_memory]               = formdata[self.user_memory][0] + 'G'
             options[self.use_jupyterlab_field]      = formdata.get(self.use_jupyterlab_field, 'unchecked')[0]
             options[self.use_local_packages_field]  = formdata.get(self.use_local_packages_field, 'unchecked')[0]
+            options[self.file]                  = formdata.get(self.file, '')[0]
```
etejedor, line 108: Can you align the `=` with the others?

SwanSpawner/swanspawner/templates/options_form_template.html
```diff
@@ -8,6 +8,19 @@
...
+            sourceForm: "lcg_options",
+            formElement: "lcg",
+        },
+        customenv: {
+            optionsElement: "builderOptions",
+            sourceForm: "customenv_options",
+            formElement: "builder",
+        }
+    };
+
```
etejedor, line 23: Do these changes have to do with launch arguments? If not, separate PR please, one commit per purpose.

SwanSpawner/swanspawner/templates/options_form_template.html
```diff
@@ -274,7 +359,7 @@
     <br><br>
     <label> User Interface <a href="#" onclick="toggle_visibility('userInterfaceDetails');"><span class='nbs'>more...</span></a></label>
     <div style="display:none;" id="userInterfaceDetails">
-        <span class='nb'>JupyterLab is the latest web-based interactive development environment for notebooks, code and data. More information <a target="_blank" href="https://jupyterlab.readthedocs.io/en/stable/user/interface.html">here</a>.</span>
+        <span class='nb'>JupyterLab is the latest web-based interactive development environment for files, code and data. More information <a target="_blank" href="https://jupyterlab.readthedocs.io/en/stable/user/interface.html">here</a>.</span>
```
etejedor, line 362: Please do not change this, it comes literally from the JupyterLab website.

SwanSpawner/swanspawner/templates/options_form_template.html
```diff
@@ -415,7 +500,7 @@ <h2>External computing resources</h2>
         <div id="clusterSection">
             <label for="clusterOptions">Spark cluster <a href="#" onclick="toggle_visibility('sparkClusterDetails');"><span class='nbs'>more...</span></a>
                 <div style="display:none;" id="sparkClusterDetails">
-                    <span class='nb'>Name of the Spark cluster to connect to from notebooks. See the <a target="_blank" href="https://hadoop-user-guide.web.cern.ch/">Hadoop User guide</a> and the <a target="_blank" href="https://sparktraining.web.cern.ch/">Spark training course</a></span>
+                    <span class='nb'>Name of the Spark cluster to connect to from files. See the <a target="_blank" href="https://hadoop-user-guide.web.cern.ch/">Hadoop User guide</a> and the <a target="_blank" href="https://sparktraining.web.cern.ch/">Spark training course</a></span>
```
etejedor, line 503: Also do not change this, we connect to the Spark clusters from notebooks.

SwanSpawner/swanspawner/templates/options_form_template.html
```diff
@@ -430,5 +515,18 @@ <h2>External computing resources</h2>
...
         </div>
+        <br>
+    </div>
+
+    <!-- File path information text -->
+    <div id="file_config" style="display: none;">
+        <h2>Open File</h2>
+        <label for="fileOption">File path <a href="#" onclick="toggle_visibility('fileDetails');"><span class='nbs'>more...</span></a>
+            <div style="display:none;" id="fileDetails">
+                <span class='nb'>Relative path to the file to be opened after the session is launched. For EOS repositories, it's relative to CERNBOX_HOME, for Git repositories, it's relative to the repository root.</span>
```
etejedor, line 526:
```suggestion
                <span class='nb'>Relative path of the file to be opened after the session is launched. For EOS repositories, it is relative to CERNBOX_HOME; for Git repositories, it is relative to the repository root.</span>
```

SwanSpawner/swanspawner/templates/options_form_template.html
```diff
@@ -248,17 +282,68 @@
...
+                set_input(elem, selected);
+            }
+            for (const [param, elem] of Object.entries(query_map.inputs)) {
+                const value = urlParams.get(param) || "";
+                set_input(elem, value);
+            }
+            if (urlParams.get('file') !== null  && urlParams.get('file') !== '') {
+                toggle_visibility('file_config');
+            }
+        } else if (software_source && !is_software_source_valid) {
```
etejedor, line 328: The second condition (`!is_software_source_valid`) is always true here, since you are the `else` of `if (is_software_source_valid)`. Therefore, no need to check it.

Regarding the first condition (`software_source`), can we just not check it and print whatever we have, even if empty? Otherwise we won't get an error message if `software_source` is empty.

In summary, this can become a single `} else {`.

SwanSpawner/swanspawner/templates/options_form_template.html
```diff
@@ -210,8 +244,8 @@
     /**
      * Adjusts the width of the repository type dropdown element based on the selected option
      */
-     function adjust_repository_dropdown() {
-        const selectElement = document.getElementById('repo_type_dropdown');
+    function adjust_repo_typeOptions() {
```
etejedor, line 247: Snake and camel case combined in a function name? Not nice...

Proposal, even if long: `adjust_repository_type_options()`.

SwanSpawner/swanspawner/templates/options_form_template.html
RodrigoSobral2000: Yes, in a way. I moved it to a bigger scope to resuse it for filling the form automatically

etejedor: Ok then.

etejedor (approved):

PR #111, opened 2024-08-07
swanspawner: redo code flow in order to prevent not provided builder arg

RodrigoSobral2000:


SwanSpawner/swanspawner/swanspawner.py
```diff
@@ -111,13 +108,17 @@ def options_from_form(self, formdata):
...
                 options[self.repo_type]       = formdata[self.repo_type][0]
-
                 if not options[self.repository]:
                     raise ValueError('Cannot create custom software environment: no repository specified')
+
+                # Builders are specified in builder-builderversion format
+                builder, builder_version = formdata[self.builder][0].lower().split('-')
+                options[self.builder]         = builder
+                options[self.builder_version] = builder_version
+                if not options[self.builder]:
```
etejedor, line 120: I believe this check is not needed, since there will be always something coming from the form. If the user filled the form from the start, it will be that value. If the user launched via URL arguments, the form will be shown anyway and the user will have to pick a value. Therefore, `builder` can't be an empty string.

PR #112, opened 2024-08-07
swanhub: new `conda` builder option

RodrigoSobral2000:
Swanhub can now also forward `conda` builder option through URL param to `SwanCustomEnvironments` extension

### DIFF
--- a/SwanHub/swanhub/spawn_handler.py
+++ b/SwanHub/swanhub/spawn_handler.py
@@ -171,10 +171,12 @@ class SpawnHandler(JHSpawnHandler):
             query_params = {
                 "repo": options.get(configs.repository),
                 "repo_type": options.get(configs.repo_type),
+                "builder": options.get(configs.builder),
                 "file": options.get(configs.file, ''),
             }
-            if options.get(configs.builder) == configs.accpy_special_type:
-                query_params[options.get(configs.builder)] = options.get(configs.builder_version)
+            # If the builder has a version, pass it as an argument of the query
+            if options.get(configs.builder_version):
+                query_params["builder_version"] = options[configs.builder_version]
 
             # Execution SwanCustomEnvs extension with the corresponding query arguments
             next_url = url_concat(url_path_join("user", user.escaped_name, "customenvs", server_name), query_params)
--- a/SwanSpawner/swanspawner/swanspawner.py
+++ b/SwanSpawner/swanspawner/swanspawner.py
@@ -113,7 +113,7 @@ def define_SwanSpawner_from(base_class):
                 if not options[self.repository]:
                     raise ValueError('Cannot create custom software environment: no repository specified')
 
-                # Builders are specified in builder:builderversion format (e.g. swan:1.0)
+                # Builders can have a version or not. When they do, we receive the following text from the form: builder:builder_version
                 raw_builder = formdata.get(self.builder, [''])[0].lower()
                 if raw_builder.count(':') == 1:
                     options[self.builder], options[self.builder_version] = raw_builder.split(':')
--- a/SwanSpawner/swanspawner/swanspawner.py
+++ b/SwanSpawner/swanspawner/swanspawner.py
@@ -113,10 +113,12 @@ def define_SwanSpawner_from(base_class):
                 if not options[self.repository]:
                     raise ValueError('Cannot create custom software environment: no repository specified')
 
-                # Builders are specified in builder-builderversion format
-                builder, builder_version = formdata[self.builder][0].lower().split('-')
-                options[self.builder]         = builder
-                options[self.builder_version] = builder_version
+                # Builders are specified in builder:builderversion format (e.g. swan:1.0)
+                raw_builder = formdata.get(self.builder, [''])[0].lower()
+                if raw_builder.count(':') == 1:
+                    options[self.builder], options[self.builder_version] = raw_builder.split(':')
+                else:
+                    options[self.builder] = raw_builder
             else:
                 options[self.lcg_rel_field]         = formdata[self.lcg_rel_field][0]
                 options[self.platform_field]        = formdata[self.platform_field][0]
--- a/SwanSpawner/swanspawner/templates/options_form_template.html
+++ b/SwanSpawner/swanspawner/templates/options_form_template.html
@@ -325,7 +325,7 @@
             if (urlParams.get('file') !== null  && urlParams.get('file') !== '') {
                 toggle_visibility('file_config');
             }
-        } else {
+        } else if (software_source !== null) {
             build_error_message(`Invalid software source: ${software_source}`);
         }
 
--- a/SwanHub/swanhub/handlers_configs.py
+++ b/SwanHub/swanhub/handlers_configs.py
@@ -37,8 +37,6 @@ class SpawnHandlersConfigs(SingletonConfigurable):
 
     customenv_special_type = 'customenv'
 
-    accpy_special_type = 'accpy'
-
     local_home = Bool(
         default_value=False,
         config=True,


SwanHub/swanhub/spawn_handler.py
```diff
@@ -173,7 +173,8 @@ def _finish_spawn(f):
                 "repo_type": options.get(configs.repo_type),
                 "file": options.get(configs.file, ''),
             }
-            if options.get(configs.builder) == configs.accpy_special_type:
+            special_builders = [configs.accpy_special_type, configs.conda_special_type]
```
etejedor, line 176: What does "special" mean here? What is the particular treatment you do for this two? This would need a comment.

RodrigoSobral2000: Means that a builder different from the default one (`python-venv`) will be sent

RodrigoSobral2000: I intend to change this logic along with the extension in the future, to make it more modular and flexible.
But for now I think is better to keep it this way

etejedor: Since we are adding conda, let's just do it the way we think it's right from the beginning, no? Why reviewing something that is not final, to do another review round afterwards?

etejedor: Also, why is the assignment below only done for conda and acc-py?

RodrigoSobral2000: I think is better to change it in a separate commit because the changes I'm planning to make will probably have further implications. Turning the PR more generic and not properly conda focused.

RodrigoSobral2000: Currently, unless a `conda` or `acc-py` is explicitly specified, the `makenv` script will build the environment with `python-venv`.

etejedor: Yes, but why `query_params[options.get(configs.builder)] = options.get(configs.builder_version)` is only done for conda and accpy?

RodrigoSobral2000: Yes, that's it.
`query_params` will be `{..., 'accpy': '2023.06', ...}` and the `makenv` script will have `--builder=builder_version`, this is `--accpy=2023.06` or `--conda=mamba`

RodrigoSobral2000: But, as mentioned, I think would be better to change the script in order to do `--builder=accpy --builder_version=2023.06` for instance

etejedor: > But, as mentioned, I think would be better to change the script in order to do --builder=accpy --builder_version=2023.06 for instance

I agree with this.

> this is --accpy=2023.06 or --conda=mamba

So the `conda` builder has a version called `mamba`? I do not think it is necessary for the conda builder to have a version?

Bottom line, query_params would container `builder_version` only in the case of accpy.

RodrigoSobral2000: I know for now either python or conda use only one "version" (venv and mamba). But for avoiding more if and elses, and even for keeping the consist format of `builder-version` on form options, I think would be better to keep this `builder_version` (and make it mandatory on the extension). This way we are also preparing ourselves for future cases like the ones we discussed about the experiments.

etejedor: You say that the venv builder also has a version called venv? But then why the assignment here is not done for venvs? I'm a bit lost.

RodrigoSobral2000: The `builder-versions` we got so far are:
- `python3-venv`
- `conda-mamba`
- `accpy-2023.06`
- `accpy-2021.12`
- `accpy-2020.11`

Im trying to follow a pattern of `<binary>-<distribution/version>`

etejedor: conda-mamba is ok, for venv I would revert it (venv-python3).

But you still haven't answered my question :) Why is the assignment only done for `special_builders` that are just conda and accpy, when you are saying that all the builders have version?

RodrigoSobral2000: Because the `python3-venv` was the default `makenv` behaviour, which means you didn't even need to specified it

RodrigoSobral2000: It was a dummy approach, but that has changed now.

etejedor: Ok. Then what I would do is not have this special variable. I would always do the assignment, with whatever comes from the form -- for accpy it will be a certain version, for venv it will always be python3, for conda it will always be mamba.

SwanSpawner/swanspawner/templates/options_form_template.html
```diff
@@ -325,7 +325,7 @@
             if (urlParams.get('file') !== null  && urlParams.get('file') !== '') {
                 toggle_visibility('file_config');
             }
-        } else {
+        } else if (software_source) {
```
etejedor, line 328: You put this back because it is necessary?

RodrigoSobral2000: Yes it is. If no `software_source` is specified, means the user is normally accessing the form, with no query arguments. So, there's no reason to show the error message. We only want to show it if the `software_source` is specified but incorrect

etejedor: Can't it happen that the user did specify a software source via URL args, but it was empty?

RodrigoSobral2000: It can, but in that case (`?software_source=`), it will be ignored

etejedor: It will be ignored and you won't notify to the user that it was wrong?

RodrigoSobral2000: This was address in a way that `?software_source=` triggers an error too, letting the user know the argument must not be empty

SwanSpawner/swanspawner/templates/options_form_template.html
```diff
@@ -325,7 +325,7 @@
             if (urlParams.get('file') !== null  && urlParams.get('file') !== '') {
                 toggle_visibility('file_config');
             }
-        } else {
+        } else if (software_source !== null) {
```
etejedor, line 328: Is this related to the conda builder options? If not, it needs to be in a separate commit, which explains why this change is needed.

SwanHub/swanhub/spawn_handler.py
```diff
@@ -171,10 +171,10 @@ def _finish_spawn(f):
             query_params = {
                 "repo": options.get(configs.repository),
                 "repo_type": options.get(configs.repo_type),
+                "builder": options.get(configs.builder),
+                "builder_version": options.get(configs.builder_version),
```
etejedor, line 175: iuuc `builder_version` will be set here as an empty string for `venv` and `mamba`?

RodrigoSobral2000: Yes, we are always sending it to the api. But it can become even more optional.

RodrigoSobral2000: Nevermind, I forgot that I have this thing of making `builder_version` optional done in a separated branch cause I was thinking about implementing it in a separated PR too. Could I include these changes in this PR (different commits tho)

etejedor: Yes please.

RodrigoSobral2000: Done, I think all changes that were needed are commited

SwanHub/swanhub/spawn_handler.py
```diff
@@ -171,10 +171,11 @@ def _finish_spawn(f):
             query_params = {
                 "repo": options.get(configs.repository),
                 "repo_type": options.get(configs.repo_type),
+                "builder": options.get(configs.builder),
                 "file": options.get(configs.file, ''),
             }
-            if options.get(configs.builder) == configs.accpy_special_type:
-                query_params[options.get(configs.builder)] = options.get(configs.builder_version)
+            if options.get(configs.builder_version):
```
etejedor, line 177: Can you add a comment here:
```
# If the builder has a version, pass it as an argument of the query
```

SwanSpawner/swanspawner/swanspawner.py
```diff
@@ -113,10 +115,13 @@ def options_from_form(self, formdata):
                 if not options[self.repository]:
                     raise ValueError('Cannot create custom software environment: no repository specified')
 
-                # Builders are specified in builder-builderversion format
-                builder, builder_version = formdata[self.builder][0].lower().split('-')
-                options[self.builder]         = builder
-                options[self.builder_version] = builder_version
+                # Builders are specified in builder:builderversion format (e.g. swan:1.0)
```
etejedor, line 118: But now you have builder and builder_options, and the latter is only set for acc-py, right? Why matching a pattern on the builder, which now can only be `venv`, `mamba` or `acc-py`?

etejedor: Ok, now I understand.

etejedor (approved):

PR #117, opened 2024-09-12
swanspawner: Verify integrity of selected options

RodrigoSobral2000:
After extracting the values from the form, the `validate_options` function takes these values and compares them to the original available ones. Only selectable options are checked in this way; free-value options are dealt with a posteriori.


SwanSpawner/swanspawner/swanspawner.py
```diff
@@ -128,8 +130,62 @@ def options_from_form(self, formdata):
 
                 self.offload = options[self.spark_cluster_field] != 'none'
 
+            self.validate_options(options)
             return options
 
+        # This function is used ensure the validity of the options selected by the user
```
etejedor, line 136: Can you write this as a function docstring ?

SwanSpawner/swanspawner/swanspawner.py
```diff
@@ -128,8 +130,63 @@ def options_from_form(self, formdata):
...
+                    if selection['type'] == 'selection' and selected_parent in [selection.get('builder', {}).get('value'), selection.get('lcg', {}).get('value')]:
+                        # The parent selection is valid and we can proceed to validate the rest of the options
+                        is_valid_parent_selection = True
+
+                        # Customenv oriented selections
+                        if self.repo_type in options and options[self.repo_type] not in self.repo_type_options:
+                            invalid_selection = self.repo_type
+
+                        # LCG oriented selections
+                        elif self.platform_field in options and options.get(self.platform_field) not in [_.get('value') for _ in selection.get('platforms')]:
```
etejedor, line 163: This should better be a generator instead of a list, to lazily check the elements.

SwanSpawner/swanspawner/swanspawner.py
```diff
@@ -128,8 +130,63 @@ def options_from_form(self, formdata):
 
                 self.offload = options[self.spark_cluster_field] != 'none'
 
+            self.validate_options(options)
             return options
 
+        def validate_options(self, options: dict):
+            """ This function is used ensure the validity of the options selected by the user """
```
etejedor, line 137:
```suggestion
            """ Ensure the validity of the options selected by the user, to prevent the acceptance of malicious / erroneous values """
```

SwanSpawner/swanspawner/swanspawner.py
```diff
@@ -166,13 +213,13 @@ def get_env(self):
                 ))
 
             # Enable JupyterLab interface
-            if self.user_options[self.use_jupyterlab_field] == 'checked':
+            if self.user_options.get(self.use_jupyterlab_field) == 'checked':
```
etejedor, line 216: This one will always be set, won't it? Why changing it to get?

SwanSpawner/swanspawner/swanspawner.py
```diff
@@ -166,13 +213,13 @@ def get_env(self):
...
                 env.update(dict(
                     SWAN_USE_JUPYTERLAB = 'true'
                 ))
 
             # Append path of user packages installed on CERNBox to PYTHONPATH
-            if self.user_options[self.use_local_packages_field] == 'checked':
+            if self.user_options.get(self.use_local_packages_field) == 'checked':
```
etejedor, line 222: Also, why use `get` here if no default option is provided anyway?

SwanSpawner/swanspawner/templates/options_form_template.html
```diff
@@ -193,7 +193,7 @@
 
     /**
      * Modifies the selection of Spark clusters and enables users to choose
-     * to use the JupyterLab interface depending on the chosen platform
+     * to use the JupyterLab interface depending on the chosen platforms
```
etejedor, line 196: No need to change this, it is still one platform selected. We only change the string literal to make it match the form, no need to update the docs.

SwanSpawner/swanspawner/swanspawner.py
rodrigo-sobral: Not if it is just assigned of lcg sessions (look at options_from_form)

SwanSpawner/swanspawner/templates/options_form_template.html
```diff
@@ -202,7 +202,7 @@
         var useLocalPackagesOption = document.getElementById('use-local-packages');
 
         /**
-         * Store the chosen platform in a hidden form field, as disabled fields
+         * Store the chosen platforms in a hidden form field, as disabled fields
```
etejedor, line 205: Same comment as above, no need to do this change.

SwanSpawner/swanspawner/swanspawner.py
rodrigo-sobral: Not if it is just assigned of lcg sessions (look at options_from_form)

SwanSpawner/swanspawner/swanspawner.py
rodrigo-sobral: For making it consistent with the others

etejedor: I'd prefer not to add more characters unnecessarily.

SwanSpawner/swanspawner/swanspawner.py
```diff
@@ -98,35 +98,82 @@ def __init__(self, **kwargs):
...
+
+                # Validate user selected options with what is on the yaml form    
+                self._validate_selection_options(selection, options)
 
                 self.offload = options[self.spark_cluster_field] != 'none'
+            else:
+                self._popup_error(options, self.software_source)
+
+            # Format resource options to do request
+            options[self.user_n_cores]  = int(options[self.user_n_cores])
```
etejedor, line 175: Extra space before the assignment, can be removed in both lines.

etejedor (approved):

PR #128, opened 2025-01-27
[SwanSpawner] Show available GPU flavours in user form

etejedor:
This is the spawner part of the series of changes to allow users to select what GPU flavour they desire in the form when they start their session.

The SwanKubeSpawner now creates a class attribute that spawns a thread. Such thread periodically updates the information about the available GPU flavours in the current kubernetes cluster. This information is now shown in the form, and is intended to be joined with the information about the current pods with GPUs attached to know if a user can get the GPU flavour they requested.

Screenshot from 2025-01-27 15-15-55


SwanSpawner/swanspawner/swankubespawner.py
```diff
@@ -21,6 +22,17 @@ class SwanKubeSpawner(define_SwanSpawner_from(KubeSpawner)):
...
     )
 
+    events_gpu_name = Unicode(
+        default_value='nvidia.com/gpu',
+        config=True,
+        help='Name in k8s of the GPU resource to be assigned to participants of an event in SWAN'
+    )
+
+    # Constant that sets a role name for participants of SWAN events
+    SWAN_EVENTS_ROLE = 'swan-events'
```
PMax5, line 32: We define the role for events in the charts (which defaults to `swan-events`): https://github.com/swan-cern/swan-charts/blob/d22d30011b2d230eb575e1e05b3e2298aeac1c0b/swan-cern/files/swan_computing_config.py#L450

If at one point we change it there, we would also need to change it here. But I assume we cannot fetch it here from the chart, right?

etejedor: As part of the related changes in the swan-charts PR:

https://github.com/swan-cern/swan-charts/pull/306/commits/066f24c0d65767b12c4c2592410fc9d35383e065#diff-e5a2b2ce9b94acef41bbd427c41215b9d473b40e87349abb9c1bc0bd9ec62d80L448-L450

I'm making that option not configurable anymore. I don't think it is that important that we can configure the label /role we use for SWAN events, so it will stay fixed, here. This has the advantage that I can access this fixed value here (spawner) and in the hooks (charts).

PMax5: Yeah I guess that we wont change it that often either

PMax5 (dismissed):

SwanSpawner/swanspawner/_gpuinfo.py
```diff
@@ -0,0 +1,176 @@
...
+import re
+from threading import Lock, Thread
+from time import sleep
+from typing import Union
+
+from kubernetes import client, config
+from kubernetes.client.models import V1NodeStatus
+from kubernetes.client.rest import ApiException
+
+class _GPUInfo:
```
diocas, line 11: `@dataclass`?

SwanSpawner/swanspawner/_gpuinfo.py
```diff
@@ -0,0 +1,176 @@
...
+            mig_config = labels['nvidia.com/mig.config']
+            if mig_config == 'disabled':
+                # Not partitioned, store info for full card
+                self._process_full_card(gpus, gpu_model, product_name, status, labels)
+            else:
+                # Partitioned, store fragment info
+                self._process_partitions(gpus, gpu_model, product_name, status)
+
+        # Fully replace the stored information (in mutual exclusion)
+        with self.get_lock():
```
diocas, line 120: Why do we need a lock if this is the only place where we update the info, the info is scoped to this class, and the function to update is called sequentially?

SwanSpawner/swanspawner/_gpuinfo.py
```diff
@@ -0,0 +1,176 @@
...
+        for node in gpu_nodes:
+            if node.spec.unschedulable:
+                # Node is cordoned, ignore its GPU
+                cordoned_gpu_nodes.append(node.metadata.name)
+                continue
+
+            status = node.status
+            labels = node.metadata.labels
+
+            product_name = labels['nvidia.com/gpu.product']
```
diocas, line 107: We can't have multiple GPUs models in a single node? I know we don't at the moment, I'm just curious.

SwanSpawner/swanspawner/_gpuinfo.py
```diff
@@ -0,0 +1,176 @@
...
+                           node_labels: dict) -> None:
+        '''
+        Gets information about allocatable GPU cards
+        '''
+        memory = int(node_labels['nvidia.com/gpu.memory']) // 1024 # GPU RAM in GB
+
+        # Check if the card is allocatable
+        resource_name = 'nvidia.com/gpu'
+        count = int(node_status.allocatable[resource_name])
+        if count > 0:
```
diocas, line 154: If we kept the option even if 0 available, would we still need to make the request that builds the full list of GPUs?

diocas: I suppose this might only be necessary for the next iteration (when we also show what is not available)

SwanSpawner/swanspawner/_gpuinfo.py
```diff
@@ -0,0 +1,176 @@
...
+                            gpus: dict,
+                            gpu_model: str,
+                            product_name: str,
+                            node_status: V1NodeStatus) -> None:
+        '''
+        Gets information about allocatable GPU partitions
+        '''
+        # Look for MIG partitions in the allocatable resources list
+        for resource_name,count in node_status.allocatable.items():
+            m = re.match('nvidia.com/mig-\d+g.(\d+)gb', resource_name) # e.g. nvidia.com/mig-1g.5gb
```
diocas, line 170: What is the 1g?

SwanSpawner/swanspawner/swankubespawner.py
```diff
@@ -21,6 +22,17 @@ class SwanKubeSpawner(define_SwanSpawner_from(KubeSpawner)):
         help='URL of the CentOS7 user image.'
     )
 
+    events_gpu_name = Unicode(
```
diocas, line 25: Is this used anywhere?

SwanSpawner/swanspawner/_gpuinfo.py
diocas: I see we use it in the spawning...

SwanSpawner/swanspawner/_gpuinfo.py
etejedor: Yes! I will change.

SwanSpawner/swanspawner/_gpuinfo.py
etejedor: We can, but I don't know how this is represented in terms of labels. I suggest we start with the assumption there is only one model per node (with possibly multiple cards of the same model) and if we ever find the other case, we update this code.

SwanSpawner/swanspawner/_gpuinfo.py
etejedor: There are two threads here:
- The thread we spawn to keep the information about the available GPU models in the cluster.
- The main thread of the hub, which spawns the user sessions.

We need a lock because we need to read/write atomically two pieces of information (_gpus and _cordoned_gpu_nodes) from two different threads.

SwanSpawner/swanspawner/_gpuinfo.py
etejedor: If the GPU is not allocatable, it does not make sense to include it in any list (it means there is some issue with it, and it can't be assigned to pods). So it's as if it did not exist.

SwanSpawner/swanspawner/_gpuinfo.py
etejedor: It represents the compute units assigned to that fragment. For the sake of simplicity, I believe it is enough to show the memory of the fragment to the user, not the compute units.

SwanSpawner/swanspawner/swankubespawner.py
etejedor: You are right, it is not. Leftover from a previous solution I tested, thanks for spotting!

SwanSpawner/swanspawner/_gpuinfo.py
diocas: Or it becomes a single pair?

SwanSpawner/swanspawner/_gpuinfo.py
diocas: It might do... You still want to show the models "installed", even if not available at the moment (i.e. that are taken).

PR #129, opened 2025-02-04
Swan on tn generate url

rodrigo-sobral:
- New button in the form for gathering all the selected options and generating a URL that fills it automatically
- New extendable section for extra options (so far only includes File Path field)

[URL here](https://rhenriqu-k8s.cern.ch/hub/spawn?software_source=customenv&use-jupyterlab=true&cores=4&memory=10&clusters=hadoop-nxcals&builder=accpy%3A2023.06&repository=https%3A%2F%2Fgitlab.cern.ch%2Frhenriqu%2Fgitlab_req&file=requirements.txt)

image

### DIFF
--- a/SwanSpawner/swanspawner/templates/options_form_template.html
+++ b/SwanSpawner/swanspawner/templates/options_form_template.html
@@ -1,6 +1,12 @@
 <style> .nb{ font-weight:normal } </style>
 <style> .nbs{ font-weight:normal; font-size:small } </style>
 <style> .news{ font-weight:normal; background-color:yellow } </style>
+<style>
+#extra_options {
+    transition: transform 0.2s ease;
+    margin-right: 5px;
+}
+</style>
 <script type="text/javascript">
 <!--
 
@@ -91,10 +97,21 @@
     }
 
     /**
+     * Rotate the triangle icon when the file options are toggled
+     */
+    function toggle_extra_options(id) {
+        const triangle = document.getElementById(id);
+        if (triangle.style.transform === "rotate(90deg)") {
+            triangle.style.transform = "rotate(0deg)";
+        } else {
+            triangle.style.transform = "rotate(90deg)";
+        }
+        toggle_visibility('fileSection');
+    }
+    /**
      * Enable display of element if previously disabled, and disable if previously enabled
      */
-    function validate_repository_input() {
-        var repoInput = document.getElementById('repositoryOption');
+    function validate_repository_input(repoInput) {
         const repoPattern = /^https?:\/\/(?:github\.com|gitlab\.cern\.ch)\/([a-zA-Z0-9_-]+)\/([a-zA-Z0-9_-]+)(\/|\.git)?$/.test(repoInput.value);
         if (repoInput.value && !repoPattern) {
             repoInput.style.backgroundColor = '#ffcccc';
@@ -323,7 +340,7 @@
                 set_input(elem, value);
             }
             if (urlParams.get('file') !== null  && urlParams.get('file') !== '') {
-                toggle_visibility('file_config');
+                toggle_extra_options('extra_options');
             }
         } else if (software_source !== null) {
             build_error_message(`Invalid software source: ${software_source}`);
@@ -460,7 +477,7 @@
                     name="repository"
                     style="flex-grow: 1; margin-left: 0; border: 1px solid #afafaf;"
                     placeholder="e.g. https://gitlab.cern.ch/user/myrepo"
-                    oninput="validate_repository_input();"
+                    oninput="validate_repository_input(this);"
                 />
             </div>
         </div>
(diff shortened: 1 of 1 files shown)


SwanSpawner/swanspawner/templates/options_form_template.html
```diff
@@ -90,11 +153,64 @@
...
+        toggle_visibility('file_config');
+    }
+
+    /**
+     * Copy the current form configuration to the clipboard
+     */
+    function copyToClipboard() {
+        const selected_source = document.querySelector('input[name="software_source"]:checked').value;
+        let generatedURL = `${window.location.href.split('?')[0]}?software_source=${selected_source}&`;
+        const selected_fields = {
```
diocas, line 175: Do we need this list of elements? Can't we understand which fields to use by their shown state? (assuming that the "extra options" options don't get removed when collapsed).

SwanSpawner/swanspawner/templates/options_form_template.html
```diff
@@ -90,11 +153,64 @@
...
+        }
+        toggle_visibility('file_config');
+    }
+
+    /**
+     * Copy the current form configuration to the clipboard
+     */
+    function copyToClipboard() {
+        const selected_source = document.querySelector('input[name="software_source"]:checked').value;
+        let generatedURL = `${window.location.href.split('?')[0]}?software_source=${selected_source}&`;
```
diocas, line 174: I'm not a fan of building these strings manually. Please use `URL` and `URLSearchParams` to construct the url and query params, instead of concatenating strings...

diocas: You can even ask the LLMs gods:

```js
const url = new URL("https://example.com");

const params = new URLSearchParams();
params.append("name", "John Doe");
params.append("age", "25");
params.append("city", "New York");

url.search = params.toString();

console.log(url.toString());
```

SwanSpawner/swanspawner/templates/options_form_template.html
```diff
@@ -90,11 +153,64 @@
...
+                generatedURL += `${param}=${value}&`;
+            }
+        }
+        if (generatedURL.endsWith('&')) {
+            generatedURL = generatedURL.slice(0, -1);
+        }
+
+        navigator.clipboard.writeText(generatedURL).then(() => {
+            const tooltip = document.querySelector('.tooltip');
+            tooltip.textContent = 'Copied!';
```
diocas, line 200: Maybe let's put it a bit clearer, e.g. "URL copied". Because the previous tooltip says "generate", and now you say "copied".. Can be confusing.

SwanSpawner/swanspawner/templates/options_form_template.html
```diff
@@ -90,11 +153,64 @@
...
+
+            setTimeout(() => {
+                tooltip.textContent = 'Generate URL';
+            }, 3000);
+        }).catch(err => {
+            console.error('Failed to copy URL: ', err);
```
diocas, line 206: Should we also show a "failed" message?

SwanSpawner/swanspawner/templates/options_form_template.html
```diff
@@ -1,6 +1,47 @@
...
+    float: right;
+    position: relative;
+    cursor: pointer;
+    width: fit-content;
+    border: 1px solid #0053A1;
+    border-radius: 8px;
+    padding: 5px;
+    transition: transform 0.2s ease, border-color 0.2s ease;
+}
+.clipboard-button img {
```
diocas, line 15: Is this still needed?

SwanSpawner/swanspawner/templates/options_form_template.html
```diff
@@ -527,9 +632,11 @@ <h2>External computing resources</h2>
         <br>
     </div>
 
+    <!-- Toggle Extra Options Section -->
+    <h2 role="button" onclick="toggle_extra_options('extra_options');" style="display: flex; align-items: center; text-decoration: none;"><span id="extra_options">></span>Extra options</h2>
```
diocas, line 636: I'm not a fan of the arrow. I would recommend using an icon instead.

diocas (changes requested):
I know you (maybe more @etejedor) don't like to hear, but I think the form is getting a bit busy... xD We should consider a re-design at some point.

SwanSpawner/swanspawner/templates/options_form_template.html
rodrigo-sobral: I don't think so. If, for some reason, it doesn't work (i don't see it happening that often), the user actively does something to make it work like refreshing the page. Showing up and error message because of this is too much, a console log should be enough for the eventuality of debugging the issue.

etejedor: @diocas do you think in this kind of case it is enough for the error to be in the console? Or we should better tell the user in the UI too?

etejedor: After talking to @diocas , he thinks showing the error as the textContent of the popup would be better. Let's implement it that way.

rodrigo-sobral: Addressed!

SwanSpawner/swanspawner/templates/options_form_template.html
```diff
@@ -90,11 +149,63 @@
...
+        }
+        toggle_visibility('fileSection');
+    }
+
+    /**
+     * Copy the current form configuration to the clipboard
+     */
+    function copyToClipboard() {
+        const selected_source = document.querySelector('input[name="software_source"]:checked').value;
+        const generatedURL = new URL(`${window.location.href.split('?')[0]}?software_source=${selected_source}&`);
```
diocas, line 170: The url contains things that shouldn't

SwanSpawner/swanspawner/templates/options_form_template.html
```diff
@@ -90,11 +149,63 @@
             e.style.display = 'block';
     }
 
+    /**
+     * Rotate the triangle icon when the file options are toggled
+     */
+    function toggle_extra_options(id) {
+        const triangle = document.getElementById(id);
+        if (triangle.style.transform === "rotate(90deg)") {
```
diocas, line 157: You might be able to do it just with css, but I won't push for it.

diocas (changes requested):

etejedor (approved):

PR #130, opened 2025-02-11
Swan on tn preventions

rodrigo-sobral:
Hide breakline properly, in order to keep the space balanced between fields
After opening a new tab on top of a running session: spawn-pending -> dialog modal to allow the user to access the current session or shut it down
If the current session is lcg: forward the user directly to the JupyterLab
If it is a customenv: forward him to the extension page in order to get the creation progress of the environment

*\*before\**
image

*\*currently\**
image

### DIFF
--- a/SwanHub/swanhub/spawn_handler.py
+++ b/SwanHub/swanhub/spawn_handler.py
@@ -43,6 +43,34 @@ class SpawnHandler(JHSpawnHandler):
         configs = SpawnHandlersConfigs.instance()
         user = self.current_user
 
+        spawner = user.get_spawner(server_name)
+
+        if spawner.ready:
+            # User has a session running, redirect to the correct page,
+            # according to the user's choice and the current session
+            spawner_software_source = spawner.user_options.get(configs.software_source)
+            if spawner_software_source == configs.customenv_special_type:
+                next_url = url_path_join("user", user.escaped_name, "customenvs", server_name)
+            else:
+                next_url = url_path_join("user", user.escaped_name, "lab")
+
+            page = await self.render_template('spawn_conflict.html', for_user=user, spawner=spawner, next_url=next_url)
+            self.finish(page)
+            return
+            
+        elif spawner.pending:
+            # If the spawner is pending, show the pending page
+            auth_state = await user.get_auth_state()
+            page = await self.render_template(
+                'spawn_pending.html',
+                for_user=user,
+                spawner=spawner,
+                progress_url=url_concat(spawner._progress_url, {"_xsrf": self.xsrf_token.decode('ascii')}),
+                auth_state=auth_state,
+            )
+            self.finish(page)
+            return
+
         # FIXME with RBAC the admin property looks like it has changed,
         # but we're going to drop this code soon either way...
         if not user.admin and os.path.isfile(configs.maintenance_file):
--- a/SwanSpawner/swanspawner/templates/options_form_template.html
+++ b/SwanSpawner/swanspawner/templates/options_form_template.html
@@ -603,7 +603,7 @@
     </div>
 
     <!-- External computing resources -->
-    <div id="external_res_config" style="display: block;">
+    <div id="external_res_config">
         <h2>External computing resources</h2>
         
         <div id="clusterSection">
@@ -628,7 +628,7 @@
     </div>
 
     <!-- Extra Options -->
-    <div id="extra_options_config" style="display: block;">
+    <div id="extra_options_config">
         <!-- Toggle Extra Options Section -->
         <h2 role="button" onclick="toggle_extra_options('extra_options');" style="display: flex; align-items: center; text-decoration: none;"><span id="extra_options">‣</span>Extra options</h2>
 
--- a/SwanSpawner/swanspawner/templates/options_form_template.html
+++ b/SwanSpawner/swanspawner/templates/options_form_template.html
@@ -623,8 +623,8 @@
                 </div>
             </label>
             <select id="condorOptions" name="condor"></select>
+            <br>
         </div>
-        <br>
     </div>
 
     <!-- Extra Options -->
--- a/SwanHub/swanhub/templates/spawn_conflict.html
+++ b/SwanHub/swanhub/templates/spawn_conflict.html
@@ -1,0 +1,57 @@
+{% extends "page.html" %}
+
+{% block main %}
+<style>
+  .btn-secondary {color: white; background-color: #fc6404;}
+  .btn-secondary:hover {color: white; background-color: #e64a00;}
+</style>
+
+<div class="container">
+  <div class="row">
+    <div class="swan-info">
+      <div id="swan-loader">
+        <div class="loader-circle">
+          <img src="{{ static_url('swan/logos/' + swan_logo_filename) }}">
+        </div>
+        <div class="loader-line-mask" hidden>
+          <div class="loader-line"></div>
+        </div>
+        <span class="text">
+          <p class="extra">
+            A session with is running already. Do you want to stop it or go to it?
+          </p>
+          <div class="extra" id="swan-options">
+            <button type="button" class="btn btn-secondary">Stop session</button>
+            <button type="button" class="btn btn-primary">JupyterLab</button>
+          </div>
+        </span>
+      </div>
+    </div>
+    <br><br>
+  </div>
+</div>
+
+{% endblock %}
+
+{% block script %}
+{{ super() }}
+<script type="text/javascript">
+require(["jquery"], function ($) {
+  $("#refresh").click(function () {
+    window.location.reload();
+  });
+  setTimeout(function () {
+    window.location.reload();
+  }, 5000);
+
+  $('#swan-options .btn-secondary').on('click', function() {
+    window.location = '{{base_url}}home?changeconfig';
+  });
+
+  $('#swan-options .btn-primary').on('click', function() {
+    window.location = '{{base_url}}'.split('/')[0] + "{{ next_url }}";
+  });
+});
+
+</script>
+{% endblock %}


SwanHub/swanhub/spawn_handler.py
```diff
@@ -43,6 +43,28 @@ async def get(self, for_user=None, server_name=''):
         configs = SpawnHandlersConfigs.instance()
         user = self.current_user
 
+        spawner = user.get_spawner(server_name, replace_failed=True)
```
etejedor, line 46: What does the `replace_failed` argument do?

SwanHub/swanhub/spawn_handler.py
```diff
@@ -43,6 +43,28 @@ async def get(self, for_user=None, server_name=''):
...
+
+        if spawner.ready:
+            # Pop-up modal for the user to choose whether to stop the session or not
+            page = await self.render_template(
+                'spawn.html', for_user=user, spawner=spawner, popup_modal='True'
```
etejedor, line 51: @diocas do you understand why we pass this field `for_user` here? @rodrigo-sobral has copied it from another call of `render_template`, in particular this:

https://github.com/swan-cern/jupyterhub-extensions/blob/master/SwanHub/swanhub/spawn_handler.py#L178

The same goes for `spawner`.

SwanHub/swanhub/spawn_handler.py
```diff
@@ -43,6 +43,28 @@ async def get(self, for_user=None, server_name=''):
...
+            )
+            self.finish(page)
+            return
+        elif spawner.pending:
+            # If the spawner is pending, show the pending page
+            auth_state = await user.get_auth_state()
+            page = await self.render_template(
```
etejedor, line 58: @diocas what do you think of this call? We took this:

https://github.com/jupyterhub/jupyterhub/blob/main/jupyterhub/handlers/pages.py#L404-L412

as an example.

What happens here is that we first show the spawn_pending page, then when the session is ready the page is reloaded, which triggers the appearance of the dialog for the user to decide whether they want to stop the session or go to it.

SwanHub/swanhub/spawn_handler.py
rodrigo-sobral:
```python
def get_spawner(self, server_name="", replace_failed=False):
        """Get a spawner by name

        replace_failed governs whether a failed spawner should be replaced
        or returned (default: returned).

        .. versionadded:: 2.2
        """
        spawner = self.spawners[server_name]
        if replace_failed and spawner._failed:
            self.log.debug(f"Discarding failed spawner {spawner._log_name}")
            # remove failed spawner, create a new one
            self.spawners.pop(server_name)
            spawner = self.spawners[server_name]
        return spawner
```
I thought i wouldn't make much sense to use it in this context, but i kept it just in case.

SwanSpawner/swanspawner/templates/options_form_template.html
```diff
@@ -603,7 +603,7 @@ <h2>Session resources</h2>
     </div>
 
     <!-- External computing resources -->
-    <div id="external_res_config" style="display: block;">
```
etejedor, line 606: Why this change and the equivalent one below? This is not related to hiding the extra line?

etejedor:
>  These changes only impact effectively lcg sessions.

This is not correct, the changes affect any get request (lcg or custom envs).

SwanHub/swanhub/spawn_handler.py
etejedor: Any argument to a function we call needs to be understood.

With the current argument, if the spawn process has failed, the failed spawner is removed from the dictionary of spawners and immediately another one is created? I don't fully understand the code, since it accesses a key from a dictionary that was just removed from that dictionary.

In any case, we need to decide if in case of failure of the spawner we want the spawner to be returned as failed or a new spawner (I would say the first, and if the spawner has failed we check it in our code and we show the form).

@diocas @rodrigo-sobral opinions please.

rodrigo-sobral: I agree, that was my intial thought. Because if the session is going to be recreated from a GET request, on top of what configuration will the session be built? As you can see by the code I sent, the spawner has a property `spawner.failed` that we can use ourselves for also preventing these cases. And if we get it, we can kill it and show the form to the user

SwanSpawner/swanspawner/templates/options_form_template.html
rodrigo-sobral: The definition of that `display: block;` is redundant (can be removed with no impact)

etejedor: But then it is not part of the same commit, isn't it? It's just a removal of redundant code that is unrelated with getting rid of the extra space?

SwanHub/swanhub/spawn_handler.py
etejedor: I assume that if `spawner.failed` is true, the hub should be already be killing the session (@diocas ?). So perhaps in that case we don't have to add any extra code and just show the form again? Or we need to show stop.html?

SwanSpawner/swanspawner/templates/options_form_template.html
rodrigo-sobral: No, the others are necessary.

SwanHub/swanhub/spawn_handler.py
```diff
@@ -47,8 +47,14 @@ async def get(self, for_user=None, server_name=''):
 
         if spawner.ready:
             # Pop-up modal for the user to choose whether to stop the session or not
+            spawner_software_source = spawner.user_options.get(configs.software_source)
+
+            next_url = url_path_join("user", user.escaped_name, "lab")
+            if spawner_software_source == configs.customenv_special_type:
+                next_url = url_path_join("user", user.escaped_name, "customenvs", server_name)
```
etejedor, line 54: Can we have an if it's custom envs we send to customenvs else we send to lab? Instead of assigning twice when it's custom envs.

etejedor: Now that I see it, these changes should be fixed up to the first commit (swanhub: Add prevention against running sessions), no need to have an extra commit that fixes things in that commit.

SwanHub/swanhub/spawn_handler.py
diocas: Maybe it's worth to test. But we might just need to show the form again, as the other tab that was trying to start already received any error it needs, no?

SwanHub/swanhub/spawn_handler.py
```diff
@@ -43,6 +43,35 @@ async def get(self, for_user=None, server_name=''):
         configs = SpawnHandlersConfigs.instance()
         user = self.current_user
 
+        spawner = user.get_spawner(server_name, replace_failed=True)
+
+        if spawner.ready:
+            # Pop-up modal for the user to choose whether to stop the session or not
+            spawner_software_source = spawner.user_options.get(configs.software_source)
+
+            if spawner_software_source == configs.customenv_special_type:
```
diocas, line 52: Do we care if we should send to customenvs or lab? I suppose we cannot assume the other tab successfully entered the customenvs page and kickstarted the env, right? If we could, then we could send the user directly to lab (since he was already informed that the session is running).

SwanHub/swanhub/spawn_handler.py
```diff
@@ -43,6 +43,35 @@ async def get(self, for_user=None, server_name=''):
...
+                next_url = url_path_join("user", user.escaped_name, "customenvs", server_name)
+            else:
+                next_url = url_path_join("user", user.escaped_name, "lab")
+            
+            page = await self.render_template(
+                'spawn.html', for_user=user, spawner=spawner, popup_modal='True', next_url=next_url
```
diocas, line 58: I would split this from the spawn page. Or is there a reason?

For consistency, this page should not look like the spawn form but like any of the other pages with the SWAN logo and the text + buttons bellow.

SwanHub/swanhub/templates/spawn.html
```diff
@@ -22,6 +22,34 @@
...
+  <div class="modal-dialog modal-dialog-centered" role="document">
+    <div class="modal-content">
+      <div class="modal-header">
+        <h4 class="modal-title" id="exampleModalLongTitle">Warning</h4>
+        <button type="button" class="close" data-dismiss="modal" aria-label="Close">
+          <span aria-hidden="true">&times;</span>
+        </button>
+      </div>
+      <div class="modal-body">
+        A session is already running. Do you want to stop it or go to it?
```
diocas, line 42: I would tell the user what changed from the current running session to what the url is asking the user to set (if it's the same or close enough, the user could decide to keep the same session).
Then would just have two buttons "Stop and create new" "Go to session"?

SwanHub/swanhub/spawn_handler.py
```diff
@@ -43,6 +43,35 @@ async def get(self, for_user=None, server_name=''):
...
+            )
+            self.finish(page)
+            return
+        elif spawner.pending:
```
diocas, line 62: Can't we keep the parent class `get` behavior instead of having this else if? I see it might require the next url (does it?), but then maybe we could fix it by overloading `get_next_url` instead? It would be cleaner and maybe could be re-used inside the previous if.

SwanHub/swanhub/spawn_handler.py
diocas: Seems fine.

SwanHub/swanhub/spawn_handler.py
etejedor: Yes I agree, we just need to show the form again. So no addition to the current code. But I would not specify `replace_failed=True` if we are unsure of what it does. If it has failed, just get it as failed, and we will show the form.

SwanHub/swanhub/spawn_handler.py
etejedor: In the case of custom envs, we can't go to lab directly because the environment might be being created, and in that case, if we went to lab, we would land in a session that does not have the environment created already (so incomplete). That's why we need this distinction, which in the case of customenvs will allow us to connect multiple tabs to an underlying script execution process that is creating the environment.

SwanHub/swanhub/templates/spawn.html
etejedor: So you want to show a diff between the args of the current session and the args of this new get? I think we can make it simpler but just saying: you either want what you are asking for or what was already there. If you don't know what was already there, then just create a session with what you want.

SwanHub/swanhub/spawn_handler.py
diocas: Yes, I would split this into a different html, as I don't see the connection between the error and this page.

Yes, but as a modal pop-up and not the error screen that we show in any other situation. Not saying that the modal isn't nice, it's just inconsistent.

SwanHub/swanhub/spawn_handler.py
diocas: I'm suggesting removing the `elif spawner.pending:` because is already checked in the parent class. So, we would call it if none of our ifs in here matched.
The only thing it seems we differ from the upstream if if that we change a bit the next url redirection. But since this is obtained by calling the method `get_next_url` we could overload it and re-use it in other places?

SwanHub/swanhub/spawn_handler.py
etejedor: I'm ok with rendering a different html for the modal, perhaps you can finish the discussion in person @diocas and @rodrigo-sobral ?

SwanHub/swanhub/spawn_handler.py
etejedor: Ok, you mean that the spawn.pending case is already handled here?

https://github.com/jupyterhub/jupyterhub/blob/main/jupyterhub/handlers/pages.py#L396-L414

The problem is that the call to the superclass `get` is only done at the end of our `get`, and before that call we render the form if we have query arguments. We would need to see how we restructure the code. Also I don't know how the `get_next_url` overload would work.

This needs to be discussed in person @diocas and @rodrigo-sobral .

SwanHub/swanhub/templates/spawn.html
etejedor: We can discuss about it, yes. @rodrigo-sobral @diocas .

SwanHub/swanhub/spawn_handler.py
```diff
@@ -43,6 +43,47 @@ async def get(self, for_user=None, server_name=''):
...
+                next_url = url_path_join("user", user.escaped_name, "lab")
+
+            query_options = {}
+            for key, byte_list in self.request.query_arguments.items():
+                query_options[key] = [bs.decode('utf8') for bs in byte_list]
+            query_options = await maybe_future(spawner.run_options_from_form(query_options))
+
+            diff_session = False
+            for config, value in query_options.items():
+                if spawner.user_options.get(config) != value and config != configs.file:
```
etejedor, line 64: For simplicity, even if optimisation is right, we agreed to remove.

SwanHub/swanhub/spawn_handler.py
```diff
@@ -43,6 +43,47 @@ async def get(self, for_user=None, server_name=''):
...
+            for key, byte_list in self.request.query_arguments.items():
+                query_options[key] = [bs.decode('utf8') for bs in byte_list]
+            query_options = await maybe_future(spawner.run_options_from_form(query_options))
+
+            diff_session = False
+            for config, value in query_options.items():
+                if spawner.user_options.get(config) != value:
+                    diff_session = True
+
+            if diff_session is True:
```
etejedor, line 67: Comment:
```
# User requests a different configuration than the current session, ask them what do to
```

SwanHub/swanhub/spawn_handler.py
```diff
@@ -43,6 +43,47 @@ async def get(self, for_user=None, server_name=''):
...
+                page = await self.render_template('spawn_conflict.html', for_user=user, spawner=spawner, next_url=next_url)
+                self.finish(page)
+            else:
```
etejedor, line 70: Comment:
```
# Same different configuration as the current session, go to it
```

etejedor (approved):

PR #134, opened 2025-03-31
swanspawner: Fix and simplify form filling with url args

rodrigo-sobral:
- Certain software stacks have their own resources and platforms. So, `adjust_form()` needs to be called for the selected lcg
- Change input background color for invalid repos in url args
- Simplify error message for invalid software source in url args
- Simplification of `if` statements

### DIFF
--- a/SwanHub/swanhub/spawn_handler.py
+++ b/SwanHub/swanhub/spawn_handler.py
@@ -87,7 +87,12 @@ class SpawnHandler(JHSpawnHandler):
         # the software_source query argument to re-render the form and fill
         # it with the remaining query arguments.
         if configs.software_source in self.request.query_arguments:
-            form = await self._render_form_wrapper(user)
+            error_message = None
+            selected_software_source = self.request.query_arguments[configs.software_source][0].decode('utf8')
+            # Check if the selected software source is valid
+            if selected_software_source not in (configs.lcg_rel_field, configs.customenv_special_type):
+                error_message = f"Invalid software source: {selected_software_source}"
+            form = await self._render_form_wrapper(user, message=error_message)
             self.finish(form)
             return
 
--- a/SwanHub/swanhub/spawn_handler.py
+++ b/SwanHub/swanhub/spawn_handler.py
@@ -52,7 +52,11 @@ class SpawnHandler(JHSpawnHandler):
             if spawner_software_source == configs.customenv_special_type:
                 next_url = url_path_join("user", user.escaped_name, "customenvs", server_name)
             else:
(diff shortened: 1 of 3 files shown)


SwanSpawner/swanspawner/templates/options_form_template.html
```diff
@@ -412,6 +412,7 @@
             for (const [param, elem] of Object.entries(query_map.selectors)) {
                 const selected = urlParams.get(param) || get_default_option(elem);
                 set_input(elem, selected);
+                if (param === 'lcg') adjust_form();
```
etejedor, line 393: I would add a comment to explain why we need this double hit.
I would also include a case for the builder.

SwanSpawner/swanspawner/templates/options_form_template.html
```diff
@@ -417,6 +417,7 @@
             for (const [param, elem] of Object.entries(query_map.inputs)) {
                 const value = urlParams.get(param) || "";
                 set_input(elem, value);
+                if (param === 'repository') validate_repository_input(document.getElementById(elem));
```
etejedor, line 398: Add comment.

SwanHub/swanhub/spawn_handler.py
```diff
@@ -52,7 +52,11 @@ async def get(self, for_user=None, server_name=''):
             if spawner_software_source == configs.customenv_special_type:
                 next_url = url_path_join("user", user.escaped_name, "customenvs", server_name)
             else:
-                next_url = url_path_join("user", user.escaped_name, "lab")
+                spawner_session_ui = spawner.user_options.get(configs.use_jupyterlab_field, None)
+                if spawner_session_ui != 'u':
```
etejedor, line 56: `!= 'u'` can also be `None`? What's the value when lab is selected?

SwanHub/swanhub/spawn_handler.py
```diff
@@ -91,7 +91,11 @@ async def get(self, for_user=None, server_name=''):
         # the software_source query argument to re-render the form and fill
         # it with the remaining query arguments.
         if configs.software_source in self.request.query_arguments:
-            form = await self._render_form_wrapper(user)
+            message = None
+            selected_software_source = self.request.query_arguments[configs.software_source][0].decode('utf8')
+            if selected_software_source not in (configs.lcg_rel_field, configs.customenv_special_type):
```
etejedor, line 96: I think these changes got mixed, they belonged to another commit, the one that removes the equivalent code in the frontend.

etejedor (approved):

PR #135, opened 2025-04-03
Display error on session start if env script path does not exist

laurall974:
Add error at the top of the user form:
Screenshot 2025-04-08 at 13 53 59

### DIFF
--- a/SwanSpawner/swanspawner/swanspawner.py
+++ b/SwanSpawner/swanspawner/swanspawner.py
@@ -200,6 +200,14 @@ def define_SwanSpawner_from(base_class):
                     value_cleaned
                 )
 
+                if int(value_cleaned) == 127:
+                    self.log.warning(
+                        "Detected user environment script setup failure (exit code 127)")
+                    raise RuntimeError(
+                        f"User environment script failed: "
+                        f"Could not find the script '{self.user_options[self.user_script_env_field]}'."
+                    )
+
             return container_exit_code
 
         async def start(self):


SwanSpawner/swanspawner/swanspawner.py
```diff
@@ -199,6 +199,9 @@ async def poll(self):
                     ".".join(["exit_container_code"]),
                     value_cleaned
                 )
+                if int(value_cleaned) == 127:
+                    self.log.warning("Detected user environment script setup failure (exit code 127)")
+                    raise RuntimeError("User environment script failed: Could not find the specified script.")
```
etejedor, line 204: You should be able to access here `self.user_options[self.user_script_env_field]` and add it to the error message, so the user has a reference of the script that we could not find.

etejedor, line 204: If you raise a RuntimeError here, I understand the user gets an error at the top the user form? Please include a screenshot in the description of the PR with the error.

etejedor, line 204: If you raise a RuntimeError here, you never return from the poll method. What happens with the user session? Is the pod terminated?

laurall974: added!

etejedor: I believe this was implemented already as per the screenshot you shared?

laurall974: I moved the code to the start method so the user session is correctly terminated.

SwanHub/swanhub/spawn_handler.py
```diff
@@ -141,7 +141,7 @@ def _finish_spawn(f):
                 user, options, time.time() - start_time_spawn, e)
 
             if type(e) in (web.HTTPError, TimeoutError):
-                error_message = configs.spawn_error_message
+                error_message = getattr(spawner, 'exit_reason', configs.spawn_error_message)
```
etejedor, line 144: I understand this commit fixes another commit in the same PR.
In a single PR, it is better not to include commits that fix other commits in the same PR. This makes it more difficult for the reviewer to review a change, because it is spread over multiple commits.
The author of the PR needs to fixup any commits that fix previous commits, and provide a clean history.

laurall974: Thanks for the tip! I just cleaned up the commit history

SwanSpawner/swanspawner/swankubespawner.py
```diff
@@ -49,6 +49,23 @@ async def start(self):
             # start configured container
             startup = await super().start()
 
+            # detect error on session start if env script path does not exist
+            for _ in range(8):
```
etejedor, line 53: Instead of iterating an arbitrary number of times until we get an exit_code, isn't there a way to run some code when the container has either successfully started or one of its containers has failed?

When does `startup = await super().start()` return? When the pod has been scheduled and its containers have started running?

etejedor: What happens if you exit 127 from the notebook container and do nothing else? Is the user pod terminated (i.e. the stop method is automatically invoked)?

laurall974: We need to wait until the timeout happens.

SwanSpawner/swanspawner/swanspawner.py
```diff
@@ -200,6 +200,14 @@ async def poll(self):
                     value_cleaned
                 )
 
+                if int(value_cleaned) == 127:
+                    self.log.warning(
+                        "Detected user environment script setup failure (exit code 127)")
+                    raise RuntimeError(
+                        f"User environment script failed: "
+                        f"Could not find or execute the script '{self.user_options[self.user_script_env_field]}'."
```
etejedor, line 208: Is it "find or execute" or just "find"?

laurall974: just "find" indeed

etejedor (approved):

PR #136, opened 2025-04-03
Fixes from Diogo suggestions

rodrigo-sobral:
Implement opening a file also for classic UI
Adapting description of file feature to lcg and custom envs
Deal with wrongly passed url args when starting a session - When a attribute is wrongly define and a session is attempted to start with it, trigger an error message that warns the user to fix the value

### DIFF
--- a/SwanHub/swanhub/templates/spawn.html
+++ b/SwanHub/swanhub/templates/spawn.html
@@ -80,7 +80,6 @@
       if (form.checkValidity()) { 
         starting = true;
         modal.modal('hide');
-        return true;
       }
     });
 
--- a/SwanHub/swanhub/templates/spawn.html
+++ b/SwanHub/swanhub/templates/spawn.html
@@ -73,10 +73,15 @@
       keyboard: false
     });
 
-    $('#spawn').on('click', function () {
-      starting = true;
-      modal.modal('hide');
-      return true;
+    $('#spawn').on('click', function (e) {
+      const form = document.getElementById('spawn_form');
+
+      // If form is valid, continue with modal hide
+      if (form.checkValidity()) { 
+        starting = true;
+        modal.modal('hide');
+        return true;
+      }
     });
 
     modal.on('shown.bs.modal', function () {
--- a/SwanSpawner/swanspawner/templates/options_form_template.html
+++ b/SwanSpawner/swanspawner/templates/options_form_template.html
@@ -376,6 +376,8 @@
         var condorSection = document.getElementById('condorSection');
         var lcg_config = document.getElementById('lcg_config');
         var jupyterLabOption = document.getElementById('use-jupyterlab');
+        var lcg_details = document.getElementById('lcg-details');
+        var customenv_details = document.getElementById('customenv-details');
 
         // Do the initial fillment of the form in order to get all the available software stacks (or builders)
         adjust_form();
@@ -414,6 +416,7 @@
             customenv_config.style.display = 'none';
             external_res_config.style.display = 'block';
             condorSection.style.display = 'block';
+            customenv_details.style.display = 'none';
             if (is_software_source_valid) {
                 adjust_options();
             } else {
@@ -421,6 +424,7 @@
             }
             adjust_platforms();
         } else {
+            lcg_details.style.display = 'none';
             document.getElementById('extra_options_config').style.display = 'block';
             previous_jupyterlab = jupyterLabOption.checked;
             adjust_clusters(external_res_config, condorSection, lcg_config, customenv_config, jupyterLabOption);
@@ -628,7 +632,8 @@
         <div id="fileSection" style="display: none;">
             <label for="fileOption">File path <a href="#" onclick="toggle_visibility('fileDetails');"><span class='nbs'>more...</span></a>
                 <div style="display:none;" id="fileDetails">
-                    <span class='nb'>Relative path of the file to be opened after the session is launched. For LCG sessions, it is relative to CERNBOX_HOME; for Custom Environments, it is relative to the specified repository root.</span>
+                    <span id="lcg-details" class='nb'>Path of the file to be opened after the session is launched, relative to CERNBOX_HOME.</span>
+                    <span id="customenv-details" class='nb'>Path of the file to be opened after the session is launched, relative to the root of the git repository.</span>
                 </div>
             </label>
             <input type="text" id="fileOption" name="file">
--- a/SwanSpawner/swanspawner/templates/options_form_template.html
+++ b/SwanSpawner/swanspawner/templates/options_form_template.html
@@ -334,6 +334,7 @@
         } else {
             platformOptions.removeAttribute('disabled');
         }
+        document.getElementById('extra_options_config').style.display = jupyterLabOption.checked ? 'block' : 'none';
     }
 
     /**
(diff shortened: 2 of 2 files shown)


SwanSpawner/swanspawner/templates/options_form_template.html
```diff
@@ -626,7 +626,7 @@ <h2 role="button" onclick="toggle_extra_options('extra_options');" style="displa
         <div id="fileSection" style="display: none;">
             <label for="fileOption">File path <a href="#" onclick="toggle_visibility('fileDetails');"><span class='nbs'>more...</span></a>
                 <div style="display:none;" id="fileDetails">
-                    <span class='nb'>Relative path of the file to be opened after the session is launched. For LCG sessions, it is relative to CERNBOX_HOME; for Custom Environments, it is relative to the specified repository root.</span>
+                    <span class='nb'>Relative path of the file to be opened after the session is launched. For LCG sessions, it is relative to CERNBOX_HOME.</span>
```
etejedor, line 629: Wait, but then, if you just get rid of that part of the message, when this section shows in the custom environment section, it will not have the right documentation.
What needs to be done is to adapt the docs depending on where you are being shown, no?

SwanHub/swanhub/spawn_handler.py
```diff
@@ -219,9 +219,13 @@ def _finish_spawn(f):
...
             next_url = url_concat(url_path_join("user", user.escaped_name, "customenvs", server_name), query_params)
         else: # LCG release
-            next_url = url_path_join(self.hub.base_url, "spawn-pending", user.escaped_name, server_name)
-            if options.get(configs.file) and options[configs.use_jupyterlab_field] == 'checked':
-                next_url = url_path_join("user", user.escaped_name, "lab", "tree", *options[configs.file].split('/'))
+            if options.get(configs.file):
+                if options[configs.use_jupyterlab_field] == 'checked':
+                    next_url = url_path_join("user", user.escaped_name, "lab", "tree", *options[configs.file].split('/'))
+                else:
+                    next_url = url_path_join("user", user.escaped_name, "edit", *options[configs.file].split('/'))
```
etejedor, line 226: This `edit` part of the URL is only there if you open e.g. a csv or txt file, but not when you open a notebook, right?

etejedor: Example of notebook URL: `https://swan-k8s.cern.ch/user/etejedor/notebooks/SWAN_projects/swanExamples2/notebooks/CppFromPython_pycpp.ipynb`
Example of csv URL: `https://swan-k8s.cern.ch/user/etejedor/edit/SWAN_projects/swanExamples2/notebooks/MuRun2010B.csv`

SwanSpawner/swanspawner/swanspawner.py
```diff
@@ -96,20 +96,27 @@ def __init__(self, **kwargs):
                 # if options_form not provided, use templated options form based on configuration file
                 self.options_form = self._render_templated_options_form
 
-        def _popup_error(self, options: dict, invalid_selection: str) -> None:
-            """ Raise an error if the selection is invalid """
-            err_msg = f'Invalid {invalid_selection} selection: {options[invalid_selection]}'
+        def _popup_error(self, err_msg: str) -> None:
```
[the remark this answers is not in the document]
rodrigo-sobral: But doing it in the frontend leaves us without any way to correctly trigger errors. Calling the `throw` javscript command only logs an error message, doesn't have practical effect. How would we trigger an error or forbid the session from being started just in the frontend?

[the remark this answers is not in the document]
etejedor: @diocas shared this link in the channel: https://developer.mozilla.org/en-US/docs/Learn_web_development/Extensions/Forms/Form_validation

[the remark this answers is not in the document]
rodrigo-sobral: I tested it. It didn't work, got completely ignored and the session got started same way.

SwanSpawner/swanspawner/templates/options_form_template.html
```diff
@@ -626,7 +627,8 @@ <h2 role="button" onclick="toggle_extra_options('extra_options');" style="displa
         <div id="fileSection" style="display: none;">
             <label for="fileOption">File path <a href="#" onclick="toggle_visibility('fileDetails');"><span class='nbs'>more...</span></a>
                 <div style="display:none;" id="fileDetails">
-                    <span class='nb'>Relative path of the file to be opened after the session is launched. For LCG sessions, it is relative to CERNBOX_HOME; for Custom Environments, it is relative to the specified repository root.</span>
+                    <span class='nb'>Relative path of the file to be opened after the session is launched. For LCG sessions, it is relative to CERNBOX_HOME; </span>
```
etejedor, line 630: Like this, when custom envs are selected, we also see the LCG description. Proposal:

LCG:
- Path of the file to be opened after the session is launched, relative to CERNBOX_HOME.
Custom env:
- Path of the file to be opened after the session is launched, relative to the root of the git repository.

SwanSpawner/swanspawner/swanspawner.py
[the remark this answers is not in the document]
etejedor: Perhaps worth investing a bit more effort?

SwanHub/swanhub/spawn_handler.py
```diff
@@ -219,9 +219,34 @@ def _finish_spawn(f):
...
+                    )
+
+                    if ext == 'ipynb':
+                        endpoint = 'notebooks'
+                    elif ext in editable_exts:
+                        endpoint = 'edit'
+                    else:
+                        # Default to /files for unknown or binary types
+                        endpoint = 'files'
+
```
etejedor, line 246: This is well implemented, but I believe it is too much, honestly. Why don't we just show the file extra option if jupyterlab is selected?

etejedor: We could do that, but then we introduce a piece of code for an interface that should be discontinued, and future SWAN admins will need to understand that they can get rid of this piece of code as soon as JLab is the only one.

I'd prefer to limit the new feature to the new interface and not leave code to clean.

SwanSpawner/swanspawner/templates/options_form_template.html
```diff
@@ -334,6 +334,7 @@
         } else {
             platformOptions.removeAttribute('disabled');
         }
+        document.getElementById('extra_options_config').style.display = jupyterLabOption.checked ? 'block' : 'none';
```
etejedor, line 337: With this, does clicking on the JupyterLab box show / hide the form?

diocas: Didn't we say that this would be ok for classic UI as well? Meaning that the change would be in the spawner next url and not here?

SwanHub/swanhub/templates/spawn.html
```diff
@@ -73,7 +73,18 @@ <h4 class="modal-title">Configure Environment</h4>
...
 
-    $('#spawn').on('click', function () {
+    $('#spawn').on('click', function (e) {
+      const form = document.getElementById('spawn_form');
+
+      if (!form.checkValidity()) {
+        // Let browser show native alerts
+        form.reportValidity();
+        // Prevent hiding the modal
+        e.preventDefault();
```
etejedor, line 83: These are all calls implemented by HTML itself?

What happens now, when a URL selects a non-valid value for an option with a restricted set of values, is it shown in red? Is the session creation blocked on the frontend? Perhaps a screenshot would help?

Also pinging @diocas .

etejedor (approved):

PR #143, opened 2025-05-05
Swan-on-TN

rodrigo-sobral:
This PR includes all the visual and logical modifications to the form, in order to configure a session using custom environemnts instead of a LCG software stack, and includes the possibility to generate URLs base on the current configuration, filling, by this way, the requirements imposed by the Swan-on-TN project.

### DIFF
--- a/SwanHub/swanhub/handlers_configs.py
+++ b/SwanHub/swanhub/handlers_configs.py
@@ -21,8 +21,6 @@ class SpawnHandlersConfigs(SingletonConfigurable):
 
     builder_version = 'builder_version'
 
-    repo_type = 'repo_type'
-
     repository = 'repository'
 
     lcg_rel_field = 'lcg'
--- a/SwanHub/swanhub/handlers_configs.py
+++ b/SwanHub/swanhub/handlers_configs.py
@@ -25,9 +25,9 @@ class SpawnHandlersConfigs(SingletonConfigurable):
 
     repository = 'repository'
 
-    lcg_rel_field = 'LCG-rel'
+    lcg_rel_field = 'lcg'
 
-    spark_cluster_field = 'spark-cluster'
+    spark_cluster_field = 'clusters'
 
     user_script_env_field = 'scriptenv'
 
--- a/SwanHub/swanhub/handlers_configs.py
+++ b/SwanHub/swanhub/handlers_configs.py
@@ -33,7 +33,15 @@ class SpawnHandlersConfigs(SingletonConfigurable):
 
     use_jupyterlab_field = 'use-jupyterlab'
 
+    use_tn = 'use-tn'
+
     customenv_special_type = 'customenv'
+
+    tn_enabled = Bool(
+        default_value=False,
+        config=True,
+        help="True if this SWAN deployment is exposed to the Technical Network."
+    )
 
     local_home = Bool(
         default_value=False,
--- a/SwanHub/swanhub/handlers_configs.py
+++ b/SwanHub/swanhub/handlers_configs.py
@@ -37,8 +37,6 @@ class SpawnHandlersConfigs(SingletonConfigurable):
 
     customenv_special_type = 'customenv'
 
-    accpy_special_type = 'accpy'
-
     local_home = Bool(
         default_value=False,
         config=True,
--- a/SwanHub/swanhub/handlers_configs.py
+++ b/SwanHub/swanhub/handlers_configs.py
@@ -33,7 +33,7 @@ class SpawnHandlersConfigs(SingletonConfigurable):
 
     use_jupyterlab_field = 'use-jupyterlab'
 
-    use_tn = 'use-tn'
+    use_tn_field = 'use-tn'
 
     customenv_special_type = 'customenv'
 
--- a/SwanHub/swanhub/handlers_configs.py
+++ b/SwanHub/swanhub/handlers_configs.py
@@ -31,6 +31,10 @@ class SpawnHandlersConfigs(SingletonConfigurable):
 
     user_script_env_field = 'scriptenv'
 
+    file = 'file'
+
+    use_jupyterlab_field = 'use-jupyterlab'
+
     customenv_special_type = 'customenv'
 
     accpy_special_type = 'accpy'
--- a/SwanSpawner/swanspawner/swankubespawner.py
+++ b/SwanSpawner/swanspawner/swankubespawner.py
@@ -50,11 +50,9 @@ class SwanKubeSpawner(define_SwanSpawner_from(KubeSpawner)):
                 raise RuntimeError('The user selected the CentOS7 platform, but no CentOS7 image was configured')
             self.image = image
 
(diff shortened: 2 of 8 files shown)


SwanHub/swanhub/handlers_configs.py
```diff
@@ -15,12 +15,34 @@ class SpawnHandlersConfigs(SingletonConfigurable):
...
+
+    spark_cluster_field = 'clusters'
 
     user_script_env_field = 'scriptenv'
 
+    file = 'file'
+
+    use_jupyterlab_field = 'use-jupyterlab'
+
+    use_tn = 'use-tn'
```
etejedor, line 36: I now realize this is not in line with the naming of the other variables, and should be `use_tn_field`?

SwanSpawner/swanspawner/swankubespawner.py
```diff
@@ -39,12 +44,22 @@ async def start(self):
...
+        if self.user_options[self.software_source] == self.lcg_special_type and 'centos7' in self.user_options[self.platform_field]:
             image = self.centos7_image
             if not image:
                 raise RuntimeError('The user selected the CentOS7 platform, but no CentOS7 image was configured')
             self.image = image
 
+        # An Accpy user image is configured by default via the chart
+        # settings. For now, we're using the accpy_image for all customenv
+        # sessions, for avoiding the installation of the extension in the swan
+        # image. This is a temporary solution.
```
etejedor, line 56: This comment is outdated.

Replace with:
```
# If the user selected an Acc-Py based custom environment,
# use the corresponding image.
```

SwanSpawner/swanspawner/swankubespawner.py
```diff
@@ -39,12 +44,22 @@ async def start(self):
...
+        if self.user_options[self.software_source] == self.customenv_special_type: # and self.user_options[self.builder] == 'accpy':
```
etejedor, line 57: We need to add the extra and condition that is commented out, since now the CustomEnvironments extension will be installed in the swan layer. This means we only create a user session with the acc-py image when they really request acc-py.

SwanSpawner/swanspawner/swanspawner.py
```diff
@@ -81,18 +108,80 @@ def __init__(self, **kwargs):
...
-            options[self.condor_pool]               = formdata[self.condor_pool][0]
-            options[self.user_n_cores]              = int(formdata[self.user_n_cores][0])
-            options[self.user_memory]               = formdata[self.user_memory][0] + 'G'
+            options[self.software_source]           = formdata[self.software_source][0]
+            options[self.user_n_cores]              = formdata[self.user_n_cores][0]
+            options[self.user_memory]               = formdata[self.user_memory][0]
+            options[self.spark_cluster_field]       = formdata.get(self.spark_cluster_field, ['none'])[0]
             options[self.use_jupyterlab_field]      = formdata.get(self.use_jupyterlab_field, 'unchecked')[0]
-            options[self.use_local_packages_field]  = formdata.get(self.use_local_packages_field, 'unchecked')[0]
+            # File to be opened when the session gets started (specified only as query parameter)
```
etejedor, line 151: It is not true anymore that it is specified only as a query parameter, right?

SwanSpawner/swanspawner/swanspawner.py
```diff
@@ -81,18 +108,80 @@ def __init__(self, **kwargs):
...
+            options[self.file]                      = formdata.get(self.file, [''])[0]
+
+            if options[self.software_source] == self.customenv_special_type:
+                options[self.repository]      = formdata[self.repository][0]
```
etejedor, line 155: We don't need the separation anymore until the assignment operator.

SwanSpawner/swanspawner/swanspawner.py
```diff
@@ -145,14 +228,16 @@ def get_env(self):
                 ))
 
             # Append path of user packages installed on CERNBox to PYTHONPATH
-            if self.user_options[self.use_local_packages_field] == 'checked':
+            if self.user_options.get(self.use_local_packages_field) == 'checked':
```
etejedor, line 231: This should only be checked for LCG, so it should be under the previous if `Enable LCG-related variables`

SwanSpawner/swanspawner/swanspawner.py
```diff
@@ -145,14 +228,16 @@ def get_env(self):
...
                 env.update(dict(
                     SWAN_USE_LOCAL_PACKAGES = 'true'
                 ))
 
             # Enable configuration for CERN HTCondor pool
-            if self.user_options[self.condor_pool] != 'none':
-                env['CERN_HTCONDOR'] = 'true'
+            if self.user_options.get(self.condor_pool, 'none') != 'none':
```
etejedor, line 237: Same here, Condor is now only possible in LCG based sessions.

SwanSpawner/swanspawner/swanspawner.py
```diff
@@ -239,22 +324,22 @@ def log_metric(self, user, host, metric, value):
...
             """
             templates_dir = os.path.join(os.path.dirname(os.path.abspath(__file__)), 'templates')
             env = Environment(loader=FileSystemLoader(templates_dir))
             template = env.get_template('options_form_template.html')
 
             try:
-                with open(self.options_form_config) as json_file:
-                    options_form_config = json.load(json_file)
+                with open(self.options_form_config) as yaml_file:
+                    options_form_config = yaml.safe_load(yaml_file)
```
etejedor, line 335: I'm confused here, are we not already using a yaml options file in master?

etejedor: https://gitlab.cern.ch/swan/gitops/-/tree/master/swan-cern/prod?ref_type=heads ---> yes we are, how come we are introducing this change only now and here?

etejedor: Or is it because the `json` library can also parse yaml?

rodrigo-sobral: Somehow, `json.load` was working even for those yaml files you mention. However, I thought i was better to use the most appropriate library and method for this purpose: `yaml.safe_load` which, according to its documentation:
```
function) def safe_load(stream: _ReadStream) -> Any
Parse the first YAML document in a stream and produce the corresponding Python object.

Resolve only basic YAML tags. This is known to be safe for untrusted input.
```

rodrigo-sobral: As you've been seeing from swan-acc-qa, it also works. But if you intend to, we can keep it as it was

etejedor: Nono, let's keep the yaml. I was just wondering why we had json in master and it worked with yaml files.

etejedor (dismissed):
Let's hold the merge for now, to be reviewed what needs to be merged together and when.

etejedor (approved):

PR #145, opened 2025-05-15
swanspawner: Uncheck checkboxes when not using alma, and fix jupyterlab selection

rodrigo-sobral:


SwanSpawner/swanspawner/templates/options_form_template.html
```diff
@@ -449,9 +450,6 @@
         setVisibleMany(['customenv_config', 'extra_options_config', 'customenv-details'], false);
         // Show LCG stack, use-local-packages, scriptenv and platforms; condor section; details about lcg in file section
         setVisibleMany(['lcg_config', 'condorSection', 'lcg-details'], true);
-
-        adjustOptions();
-        adjustPlatforms();
```
etejedor, line 450: Are these removals related to this change?

SwanSpawner/swanspawner/templates/options_form_template.html
```diff
@@ -306,8 +306,9 @@
 
         // Hide the extra options section if JupyterLab is not selected
         setVisible('extra_options_config', jupyter_lab_option.checked);
-        // Remember last value of jupyterlab checkbox option each time it gets changed
-        previous_jupyterlab = jupyter_lab_option.checked;
```
etejedor, line 306: I'm having second thoughts about this... perhaps it's just simpler to keep the "use jupyterlab" value always, independently of how it was set (explicitly by the user or by selecting custom envs). Otherwise we are filling this logic with cornercases.

SwanSpawner/swanspawner/templates/options_form_template.html
```diff
@@ -155,6 +155,8 @@
     function setVisible(id, visible) {
         const elem = document.getElementById(id);
         elem.style.display = visible ? 'block' : 'none';
+        // Enable all inputs if the element is visible, and disable them if it is not visible
```
etejedor, line 158: You it is not possible just to set `disabled` on this `elem` for which we change the visibility? And that affects everything within it in the same way that `elem.style.display = value` does for visibility?

SwanSpawner/swanspawner/templates/options_form_template.html
etejedor: Let's just go ahead with this, it improves usability.

SwanSpawner/swanspawner/templates/options_form_template.html
```diff
@@ -292,7 +296,7 @@
      * Disable the platform options if the user selects JupyterLab or local packages
      * and show/hide the extra options section
      */
-    function adjustPlatforms() {
+    function adjustPlatforms(by_click) {
```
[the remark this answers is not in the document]
rodrigo-sobral: This is simply a good practice in general code. So, the code doesn't fail if the argument is not specifed (but also hard to debug). But it can be implemented for making sure we work only with booleans

PR #146, opened 2025-05-16
swanspawner: Take current form when switching clusters

rodrigo-sobral:


SwanSpawner/swanspawner/templates/options_form_template.html
```diff
@@ -494,22 +494,29 @@
             // the form won't get submitted, so they need to be filled with the default values
             adjustForm(lcg_source);
             // The customenv form needs to be initially filled in order to preserve the main dropdown and extract its options as well
-            adjustForm(customenv_source);
+            if (enable_customenv_form === "true") adjustForm(customenv_source);
```
etejedor, line 506: I think the commit message should be `Do not fill or show a customenv form if `enable_custom…

…env_form` is false`?

Anyway, do we need this logic? Can't we just ignore if the customenv form is enabled or not? I know we will fill it every time even if it's disabled,  but do we care? And we would not need this extra logic.

rodrigo-sobral: Right, my mistake, thx.

Not really, the two `showLCGForm` and `showCustomEnvForm` functions are the ones that make the management of elements between forms, which means that if you don't call any of them. You'll see all fields regarding lcg and customenv sessions.
This logic behind not running `adjustForm` for customenvs in this case, is made not only in order to avoid running unnecessary code and optimising its performance, but also because we cannot do the initial filling of the customenv form, since we don't have the `customenv_options` specifed in the `values_options_form.yaml`.

etejedor: But if we don't have the customenv_options in the values_options_form.yaml of the general SWAN, can't we just detect that and therefore not fill it?

SwanHub/swanhub/templates/spawn.html
```diff
@@ -74,7 +74,12 @@ <h4 class="modal-title">Configure Environment</h4>
     });
 
     // Fill the checkbox if this is a cluster exposed to the TN
-    $('#use-tn').prop('checked', '{{ tn_enabled }}' === 'True');
+    window.onpageshow = (event) => {
+      if (event.persisted) {
```
etejedor, line 78: What does this mean?

etejedor: Ah, I think I know per the commit message. This is to ensure that when we show the value of tn access, even when going back with the browser, it is always consistent with the instance it is running (normal or ats).

rodrigo-sobral: Exactly, that's it

PR #151, opened 2025-05-30
swan culler: implement mechanism to delete sessions of blocked users

laurall974:
this implements the new logic in the culler hub service that detects, every 12 hours, which users have been blocked or disabled. For blocked users, their entire session is terminated, thus preventing any surviving Websocket to allow them to execute commands in their session.


SwanCuller/swanculler/app.py
```diff
@@ -414,6 +537,24 @@ def main():
...
+        auth_url = options.auth_url,
+        client_id=options.auth_client_id,
+        client_secret=options.auth_client_secret,
+        audience = options.audience,
+        authz_api_url=options.authz_api_url,
+    )
+    # schedule first cull immediately
+    # because PeriodicCallback doesn't start until the end of the first interval
+    loop.add_callback(blocked_check)
+    # schedule periodic cull
```
etejedor, line 554: Comment not correct, it's not culling.

SwanCuller/swanculler/app.py
```diff
@@ -381,6 +498,12 @@ def main():
     )
     define('hooks_dir', default="/srv/jupyterhub/culler", help="Path to the directory for the krb tickets script (check_ticket.sh)")
     define('disable_hooks', default=False, help="The user's home is a temporary scratch directory and we should not check krb tickets")
+    define('auth_url', default='', help="URL to fetch CERN access token")
+    define('auth_client_id', default=os.environ.get('AUTH_CLIENT_ID'), help="Client ID for blocked user check")
+    define('auth_client_secret', default=os.environ.get('AUTH_CLIENT_SECRET'), help="Client secret for blocked user check")
```
etejedor, line 503: These two you need to add to sops?

SwanCuller/swanculler/app.py
```diff
@@ -346,6 +346,123 @@ def handle_user(user):
...
+        except HTTPClientError as e:
+            if e.code == 404:
+                app_log.info(f"User {user['name']} not found for blocked user check, skipping.")
+                continue
+            else:
+                app_log.warning(f"Failed to check identity for {user['name']}: {e}")
+                continue
+
+        if is_blocked or not is_active:
+            app_log.warning("User %s is blocked or disabled. Terminating their sessions.", user['name'])
```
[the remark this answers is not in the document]
laurall974: According to the authorization API doc:

- blocked: True if the resource is blocked.
- activeUser: Flag indicating if the person has an active affiliation with CERN.

So I assumed that the first case refers to a restricted account and the second one to an "expired" account and in both cases session should be terminated?

SwanCuller/swanculler/app.py
```diff
@@ -346,6 +346,123 @@ def handle_user(user):
...
+            for i, server_name in enumerate(servers):
+                try:
+                    resp = results[i]
+                    if resp.code in (204, 202):
+                        app_log.info("Deleted server '%s' for user %s", server_name, user['name'])
+                    else:
+                        app_log.warning("Unexpected response deleting server %s for user %s: %s",
+                                        server_name, user['name'], resp.code)
+                except Exception:
+                    app_log.exception("Failed to delete server %s for user %s", server_name, user['name'])
```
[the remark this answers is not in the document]
laurall974: Updated!

SwanCuller/swanculler/app.py
etejedor: Can you point to the line where this happens? I would assume these are secrets stored in sops?

laurall974: It is define here:
- https://github.com/swan-cern/swan-charts/blob/40b85d8de85b4af1ca5a66a4cde63e3756152d8a/swan-cern/values.yaml#L177C1-L178C1 
- https://github.com/swan-cern/swan-charts/blob/40b85d8de85b4af1ca5a66a4cde63e3756152d8a/swan/values.yaml#L157
and then in this PR I used it as "c.KeyCloakAuthenticator.client_id": 
https://github.com/swan-cern/swan-charts/pull/366/files

etejedor: Oh ok so those are preexisting secrets we already have in production, I see.

etejedor (approved):

PR #154, opened 2025-06-04
swanhub: Append `use-jupyterlab` value to the `next_url` call

rodrigo-sobral:
This change makes the `SwanContents` extension able to open the downloaded project either on ClassicUI or JupyterLab, depending on the given checkbox value

### DIFF
--- a/SwanHub/swanhub/spawn_handler.py
+++ b/SwanHub/swanhub/spawn_handler.py
@@ -217,8 +217,13 @@ class SpawnHandler(JHSpawnHandler):
             next_url = url_concat(url_path_join("user", user.escaped_name, "customenvs", server_name), query_params)
         else: # LCG release
             next_url = self.get_next_url(user, default=url_path_join(self.hub.base_url, "spawn-pending", user.escaped_name, server_name))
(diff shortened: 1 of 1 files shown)


SwanHub/swanhub/spawn_handler.py
```diff
@@ -217,8 +217,11 @@ def _finish_spawn(f):
             next_url = url_concat(url_path_join("user", user.escaped_name, "customenvs", server_name), query_params)
         else: # LCG release
             next_url = self.get_next_url(user, default=url_path_join(self.hub.base_url, "spawn-pending", user.escaped_name, server_name))
-            if options.get(configs.file) and options[configs.use_jupyterlab_field] == 'checked':
-                next_url = url_path_join("user", user.escaped_name, "lab", "tree", *options[configs.file].split('/'))
+            if options[configs.use_jupyterlab_field] == 'checked':
+                if 'next' in self.request.query_arguments:
```
etejedor, line 221:
```suggestion
                if 'next' in self.request.query_arguments: # Open in SWAN (we have "next" argument)
```

PR #159, opened 2025-06-18
SwanSpawner: Implement a mechanism to update information on free GPUs

laurall974:
This implement a mechanism which consists in storing the information about free GPU models at pod creation AND deletion time, with the ultimate goal to show to the user only the GPU models that are free.The GPU models that are not free are still be shown in the form, but greyed out.

Screenshot 2025-06-18 at 13 44 17
Screenshot 2025-06-18 at 08 47 02


SwanSpawner/swanspawner/_gpuinfo.py
```diff
@@ -81,6 +90,73 @@ def _update_gpu_info(self) -> None:
         while True:
             self._update_allocatable_gpu_flavours()
             sleep(self.UPDATE_INTERVAL)
+    
+    def _watch_pod_events(self) -> None:
+        '''
+        Watches for pod creation and deletion events in the SWAN namespace that involve GPU resources.
+        '''
+        w = watch.Watch()
+        while not self._stop_event.is_set():
```
etejedor, line 99: I remember using this `watch` API. The problem I saw is that, even though you set a stream of events for namespaced pods, the number of (uninteresting events) for us was high. So we had this thread in a busy wait in a loop where maybe 95 % of the iterations were useless. Can you confirm this?

laurall974: Updated to not used watch anymore.

SwanSpawner/swanspawner/_gpuinfo.py
```diff
@@ -20,6 +20,7 @@ class _GPUInfo:
     resource_name: str
     product_name: str
     count: int = 0
+    available: int = 0
```
etejedor, line 23: Change to `free`? And please document in the docstring above. To make the difference clear, `count` could be documented as number of available, and `free` as number of currently free.

SwanSpawner/swanspawner/_gpuinfo.py
```diff
@@ -41,6 +44,70 @@ def __init__(self, events_role: str):
...
+        '''
+        try:
+            # Get all running pods in the namespace that request GPU
+            all_pods = self._api.list_namespaced_pod(
+                namespace=self.NAMESPACE,
+                field_selector='status.phase=Running'
+            ).items
+            gpu_usage_by_resource = {}
+
+            for pod in all_pods:
```
etejedor, line 60: `for pod in user_gpu_pods`, if you filter as suggested earlier.

SwanSpawner/swanspawner/swankubespawner.py
```diff
@@ -108,12 +108,31 @@ async def stop(self, now=False):
...
                         self.log.error('Error deleting secret {namespace}:{hadoop_secret_name}: {e}')
 
+            # GPU availability update
+            try:
+                gpu_flavour = self.user_options.get('gpu')
+                if gpu_flavour:
+                    await self.gpus.increment_gpu_availability(gpu_flavour)
+                    self.log.info(f"Incremented availability for GPU {gpu_flavour} after user stop.")
+            except Exception as e:
+                self.log.error(f"Failed to increment GPU availability: {e}")
```
etejedor, line 118: Also here, I believe we should not do a simple decrement of that particular model, I believe we should also take this chance to query all gpu pods and update the whole database. If I understand correctly (please check) this code here runs when the pod has already been terminated. This means that if you query and update the database, that's the good status already, you don't need to decrement.

SwanSpawner/swanspawner/swankubespawner.py
```diff
@@ -108,12 +108,31 @@ async def stop(self, now=False):
...
-        self._dynamic_form_info['gpu_flavours'] = sorted(gpu_flavours, key=lambda x: gpu_flavours.get(x).count, reverse=True)
+        gpu_flavours = []
+        for flavour, gpu_info in gpu_flavours_raw.items():
+            gpu_flavours.append({
+                'name': flavour,
+                'available': gpu_info.available,
+                'total': gpu_info.count,
+                'resource_name': gpu_info.resource_name,
+                'enabled': gpu_info.available > 0
+            })
```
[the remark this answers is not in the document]
etejedor: You can just invoke a backend function (e.g. self.gpus.get_free_flavours()) that gets you the flavors with `free` > 0, and you pass that.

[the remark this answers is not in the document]
laurall974: I modified it, I thought it would be interesting to see the available GPU on the cluster in general.

SwanSpawner/swanspawner/templates/options_form_template.html
```diff
@@ -403,36 +452,17 @@
                 
                 // Enable GPU selection only if CUDA stack is used
                 if (selectLCG.selectedOptions[0] && selectLCG.selectedOptions[0].value.includes('cuda')) {
-                    var gpuFlavours = dynamicFormInfo['gpu_flavours'];
```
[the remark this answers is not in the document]
laurall974: I put both so we can still show all available flavor and greyed out the on that are not free as requested in the Jira ticket.

SwanSpawner/swanspawner/swanspawner.py
```diff
@@ -156,8 +156,9 @@ def options_from_form(self, formdata: dict) -> dict:
             options[self.lcg_rel_field]             = formdata[self.lcg_rel_field][0]
             options[self.platform_field]            = formdata[self.platform_field][0]
             options[self.user_script_env_field]     = formdata[self.user_script_env_field][0]
-            options[self.spark_cluster_field]       = formdata[self.spark_cluster_field][0] if self.spark_cluster_field in formdata.keys() else 'none'
+            options[self.spark_cluster_field]       = formdata.get(self.spark_cluster_field, ['none'])[0]
```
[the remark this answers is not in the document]
laurall974: I assume it was a merge error, from the previous code version and the merge from the TN. Just removed it.

SwanSpawner/swanspawner/swanspawner.py
```diff
@@ -156,8 +156,9 @@ def options_from_form(self, formdata: dict) -> dict:
...
             options[self.condor_pool]               = formdata[self.condor_pool][0]
+            options[self.software_source]           = formdata[self.software_source][0]
```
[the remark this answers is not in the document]
laurall974: Same comment as above.

SwanSpawner/swanspawner/_gpuinfo.py
```diff
@@ -126,6 +190,8 @@ def _update_allocatable_gpu_flavours(self) -> None:
                           f'{[(flavour,gpu_info.count) for flavour,gpu_info in self._gpus.items()]}')
         self._logger.info(f'Cordoned GPU nodes: {self._cordoned_gpu_nodes}')
 
+        self._check_free_gpu()
```
etejedor, line 193: Update docstring of function to state that free gpus are also updated.

etejedor, line 193: Move up as a single call outside of this function.

SwanSpawner/swanspawner/_gpuinfo.py
```diff
@@ -1,12 +1,12 @@
 from dataclasses import dataclass
 import logging
 import re
-from threading import Lock, Thread
-from time import sleep
+from threading import Lock, Thread, Event
+from time import sleep, time
```
etejedor, line 5: Unused import 2.

SwanSpawner/swanspawner/_gpuinfo.py
```diff
@@ -16,19 +16,24 @@ class _GPUInfo:
...
     '''
     Class that spawns a thread that updates the information about
     available GPU flavours in the k8s cluster at regular intervals.
     A GPU flavour can correspond to a GPU model (e.g. "Tesla T4") or to
     a partition of a GPU card (e.g. "A100 fragment (5 GB)" ).
+    It also check how many GPUs are currently free out of the available ones.
     '''
     UPDATE_INTERVAL = 60 * 10  # 10 minutes
+    NAMESPACE = 'swan'
+    CONTAINER_NAME = 'notebook'
```
etejedor, line 36: Unused variable.

SwanSpawner/swanspawner/_gpuinfo.py
```diff
@@ -41,6 +46,52 @@ def __init__(self, events_role: str):
...
+            gpu_usage_by_resource = {}
+
+            for pod in user_gpu_pods:
+                if pod.spec.node_name in self._cordoned_gpu_nodes:
+                    continue
+                container = next((c for c in pod.spec.containers if c.name == 'notebook'), None)
+                for resource_name, quantity in container.resources.requests.items():
+                    if resource_name.startswith('nvidia.com/'):
+                        try:
+                            gpu_count = int(quantity)
```
etejedor, line 69: To be clear, `requested_gpu_count`?

SwanSpawner/swanspawner/_gpuinfo.py
```diff
@@ -41,6 +46,52 @@ def __init__(self, events_role: str):
...
+                            )
+                        except ValueError:
+                            self._logger.warning(
+                                f'Could not parse GPU quantity "{quantity}" for resource "{resource_name}" '
+                                f'in pod {pod.metadata.name}, container {container.name}'
+                            )
+            # Update free GPU counts
+            with self._lock:
+                for description, gpu_info in self._gpus.items():
+                    used_count = gpu_usage_by_resource.get(gpu_info.resource_name, 0)
```
etejedor, line 81: Hehe, here you actually changed the name to something similar to what I suggested. I would make the naming consistent above and here.

SwanSpawner/swanspawner/_gpuinfo.py
```diff
@@ -41,6 +46,52 @@ def __init__(self, events_role: str):
...
+                container = next((c for c in pod.spec.containers if c.name == 'notebook'), None)
+                for resource_name, quantity in container.resources.requests.items():
+                    if resource_name.startswith('nvidia.com/'):
+                        try:
+                            gpu_count = int(quantity)
+                            gpu_usage_by_resource[resource_name] = (
+                                gpu_usage_by_resource.get(resource_name, 0) + gpu_count
+                            )
+                        except ValueError:
+                            self._logger.warning(
```
etejedor, line 74: This should be logged as an error, no? For me it's at the same level as the errors you log below.

SwanSpawner/swanspawner/swankubespawner.py
```diff
@@ -108,12 +108,23 @@ async def stop(self, now=False):
                     except ApiException as e:
                         self.log.error('Error deleting secret {namespace}:{hadoop_secret_name}: {e}')
 
+            # free GPU update
+            try:
+                gpu_flavour = self.user_options.get('gpu')
+                if gpu_flavour:
+                    self.gpus._check_free_gpu()
```
etejedor, line 115: This is the only line that can throw, right? Then I would put the lines before and after outside of the try block -- it's good to have thin try blocks to be able to identify faster what can throw.

SwanSpawner/swanspawner/swanspawner.py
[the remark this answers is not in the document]
etejedor: Sorry I don't see anymore the comment above :) Can you repeat?

SwanSpawner/swanspawner/templates/options_form_template.html
```diff
@@ -403,36 +403,39 @@
...
-                    
+                    const gpuFlavours = dynamicFormInfo['gpu_flavours'] || [];
```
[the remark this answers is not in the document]
laurall974: yes! the gpu field became madantory, so you can't submit if it set to None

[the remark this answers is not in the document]
etejedor: But it can still be None and submit when the software stack is non gpu, right? We just want to block the submission if the software stack is gpu and there's no gpu selected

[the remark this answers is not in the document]
laurall974: Now the gpu is required only if cuda is selected already, otherwise you can submit normally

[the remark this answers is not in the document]
etejedor: You mean that when a non cuda stack is selected, the form entry where the GPU is selected is disabled? And therefore the form can be submitted?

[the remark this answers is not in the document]
laurall974: Yes exactly!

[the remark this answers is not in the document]
etejedor: Ok then we just have to figure out the order in which the flavours are shown (and therefore the flavour that is selected by default, the one at the top of the list).

SwanSpawner/swanspawner/_gpuinfo.py
```diff
@@ -6,7 +6,7 @@
 from typing import Union
 
 from kubernetes import client, config
-from kubernetes.client.models import V1NodeStatus
+from kubernetes.client.models import V1NodeStatus, V1Pod
```
etejedor, line 9: Import not used.

etejedor (approved):

PR #172, opened 2025-09-18
swanculler: Add global list of users for the culler and blocked_users functions

PMax5:
This solution intends to update a single list of users to be used by the function responsible to cull user sessions and the function that validates the status of SWAN users.

It intends to fix the issue which prevents fetching the list of users on the blocked users function

### DIFF
--- a/SwanCuller/swanculler/app.py
+++ b/SwanCuller/swanculler/app.py
@@ -55,6 +55,11 @@ from tornado.options import define, options, parse_command_line
 # Start SWAN code
 from subprocess import call
 
+# Global variables
+# The users list is shared between the functions that check for idle
+# sessions and the one that validates wether a user is blocked
+users = []
+
 def check_ticket(username):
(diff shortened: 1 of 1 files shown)


SwanCuller/swanculler/app.py
```diff
@@ -143,7 +148,11 @@ def fetch(req):
         fetch = client.fetch
 
     resp = yield fetch(req)
-    users = json.loads(resp.body.decode('utf8', 'replace'))
+
+    users_list = json.loads(resp.body.decode('utf8', 'replace'))
+
+    users.clear()
+    users.extend(users_list)
```
etejedor, line 155: Why doing all this instead of just rebinding the variable (and letting the previous value be garbage collected)?

PMax5: Yeah that is a better approach indeed. Changed!

etejedor (approved):

PR #212, opened 2026-01-26
KeyCloakAuthenticator: check for "use" == "sig" when discovering signature public key from provider

jezwilkinson:
Fixes an issue in the OIDC signature public key discovery in KeyCloakAuthenticator, in cases where the KeyCloak realm provides multiple keys at its OIDC `/certs` endpoint.

Current behaviour: If `c.KeyCloakAuthenticator.check_signature` is set to True, the authenticator polls the "jwks_uri" endpoint (realm/protocol/openid-connect/certs) during startup and returns key[0] as the public key to use for verifying token signatures. This works fine if the signing key is the first or only one provided at the `certs` endpoint, but leads to signature validation errors whenever a user logs in if the login realm's `/certs` endpoint provides a different public key (e.g. one used for encryption) in the first position of the JWK object.

This commit allows KeyCloakAuthenticator to automatically determine which of the public keys on the `certs` endpoint is designated as the signing key, by checking [the "use" parameter](https://datatracker.ietf.org/doc/html/rfc7517#section-4.2) of each key, and returns the topmost key marked as a signature verification key (use == "sig" ). Additionally, if no "sig" key is found, then an exception is raised.


rodrigo-sobral (dismissed):

PMax5 (approved):

tomasr8 (approved):

tomasr8:
Thanks @jezwilkinson!

PR #244, opened 2026-04-17
swanhub: Log when scriptenv is used

tomasr8:
It's currently completely skipped. We can simply log whether it was used or not without logging the actual value.

### DIFF
--- a/SwanHub/swanhub/spawn_handler.py
+++ b/SwanHub/swanhub/spawn_handler.py
@@ -287,11 +287,12 @@ class SpawnHandler(JHSpawnHandler):
         configs = SpawnHandlersConfigs.instance()
 
         for (key, value) in options.items():
-            if key != configs.user_script_env_field:
+            if key == configs.user_script_env_field:
+                # For the environment script, we want to log only whether it is set or not, not the actual value
+                value_cleaned = 'set' if value else 'not_set'
+            else:
                 value_cleaned = str(value).replace('/', '_')
-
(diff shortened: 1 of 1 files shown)


SwanHub/swanhub/spawn_handler.py
```diff
@@ -292,6 +292,10 @@ def _log_spawn_metrics(self, user, options, spawn_duration_sec, spawn_exception=
 
                 self._log_metric(user.name, host, ".".join(
                     ['spawn_form', key]), value_cleaned)
+            else:
+                # For the environment script, we want to log only whether it is set or not, not the actual value
+                self._log_metric(user.name, host, ".".join(
+                    ['spawn_form', key]), 'set' if value else 'not_set')
```
tomasr8: Not sure which set of values would be best here? `set/not_set`, `true/false`, or something else?

rodrigo-sobral: I'd suggest to do something like this:
```python
            if key == configs.user_script_env_field:
                # For the environment script, we want to log only whether it is set or not, not the actual value
                value_cleaned = 'enabled' if value else 'disabled'
            else:
                value_cleaned = str(value).replace('/', '_')

            self._log_metric(user.name, host, ".".join(['spawn_form', key]), value_cleaned)
```

tomasr8: Ah yeah, that makes it simpler! Updated

rodrigo-sobral (approved):

PR #288, opened 2026-07-06
KeyCloakAuthenticator:implement tests for _get_oidc_configs function

icefox97:
includes tests for all branches in the _get_oidc_configs function, implements the unconfigured_authenticator monkeypatch to break out of loop and replace sleep.

branch options are:
- authorize_url, token_url, userdata_url not all present, expect exception and have to break loop by patching sleep                                                
 [test_missing_required_authorisation_fields]

- authorize_url, token_url, userdata_url all present, 
  - enable_logout and end_session present [test_set_urls]
      - logout_redirect_url not present [test_enable_logout_end_session_no_existing_redirect_url]
      - logout_redirect_url present [test_enable_logout_end_session_existing_redirect_url]
  - enable_logout and end_session not all present [test_logout_url_not_updated]
  
- check_signature True
    - sign_keys True [test_check_signature_true_sig_key_present]
    - sign_keys False [test_check_signature_true_no_sig_key]
   
 - check_signature False [test_check_signature_false_skip_jwks_fetch]

### DIFF
--- a/KeyCloakAuthenticator/keycloakauthenticator/tests/test_auth.py
+++ b/KeyCloakAuthenticator/keycloakauthenticator/tests/test_auth.py
@@ -229,6 +229,23 @@ class TestGetOidcConfigs:
 
         assert unconfigured_authenticator.public_key is not None
 
+    
+    @pytest.mark.asyncio
+    async def test_check_signature_false_skip_jwks_fetch(self, unconfigured_authenticator, monkeypatch):
+        call_count = 0
+
+        async def mock_httpfetch(url, **kwargs):
+            nonlocal call_count
+            call_count += 1
+            return OIDC_DISCOVERY_DOC
+
+        monkeypatch.setattr(unconfigured_authenticator, "httpfetch", mock_httpfetch)
+
+        await unconfigured_authenticator._get_oidc_configs()
+
+        assert call_count == 1
+        assert unconfigured_authenticator.public_key is None
+
 
 
 @pytest.mark.asyncio
--- a/KeyCloakAuthenticator/keycloakauthenticator/tests/test_auth.py
+++ b/KeyCloakAuthenticator/keycloakauthenticator/tests/test_auth.py
@@ -1,4 +1,5 @@
 import json
+import asyncio
 
 import jwt
 import pytest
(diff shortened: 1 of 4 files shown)


KeyCloakAuthenticator/keycloakauthenticator/tests/test_auth.py
```diff
@@ -32,6 +34,220 @@ def _get_mock_token(private_key, token_id, expired=False):
...
+
+        - enable_logout and end_session not all present
+
+        - check_signature True
+            - sign_keys True
+            - sign_keys False
+        - check_signature False
+    """
+
+    @pytest.mark.asyncio
```
tomasr8: Can we use `asyncio_mode = auto` instead of marking every method? https://pytest-asyncio.readthedocs.io/en/latest/reference/configuration.html#asyncio-mode

KeyCloakAuthenticator/keycloakauthenticator/tests/test_auth.py
```diff
@@ -32,6 +34,220 @@ def _get_mock_token(private_key, token_id, expired=False):
...
+        {"authorization_endpoint": "http://fake/auth"},
+        {"authorization_endpoint": "http://fake/auth", "token_endpoint": "http://fake/token"},
+        {"token_endpoint": "http://fake/token"},
+    ])
+    async def test_missing_required_authorisation_fields(self, unconfigured_authenticator, monkeypatch, doc):
+        async def mock_httpfetch(url, **kwargs):
+            return doc
+
+        async def mock_sleep(_):
+            raise RuntimeError("breaking retry loop")
```
tomasr8: Hmm the infinite loop combined with the catch-all except in the `_get_oidc_configs` method is pretty nasty. Especially the fact that it hides the exceptions being raised in the inner loop.

I wonder if we could do a small refactor where we take everything inside the `try` block and put it in a separate method. Then we'll be able to write tests without the monkeypatching and we can also assert the exact exception we're expecting to get. For example in this case, we'd be able to assert `Unable to retrieve OIDC necessary values` directly.

KeyCloakAuthenticator/keycloakauthenticator/tests/test_auth.py
```diff
@@ -32,7 +34,239 @@ def _get_mock_token(private_key, token_id, expired=False):
...
+
+def _make_jwks(public_key, use_sig=True):
+    jwk = json.loads(RSAAlgorithm.to_jwk(public_key))
+    if use_sig:
+        jwk["use"] = "sig"
+    return {"keys": [jwk]}
+
+# ---------------------------------------------------------------------------
+# TestGetOidcConfigs
+# ---------------------------------------------------------------------------
```
tomasr8: Let's get rid of these, it's just extra noise
```suggestion
```

KeyCloakAuthenticator/keycloakauthenticator/tests/test_auth.py
```diff
@@ -32,7 +34,239 @@ def _get_mock_token(private_key, token_id, expired=False):
...
+
+OIDC_DISCOVERY_DOC = {
+    "authorization_endpoint": "http://fake/auth",
+    "token_endpoint": "http://fake/token",
+    "userinfo_endpoint": "http://fake/userinfo",
+}
+
+
+def _make_jwks(public_key, use_sig=True):
```
tomasr8:
```suggestion
def _make_jwks(public_key, *, use_sig=True):
```

KeyCloakAuthenticator/keycloakauthenticator/tests/test_auth.py
```diff
@@ -32,7 +34,239 @@ def _get_mock_token(private_key, token_id, expired=False):
...
+            - logout_redirect_url not present
+            - logout_redirect_url present
+
+        - enable_logout and end_session not all present
+
+        - check_signature True
+            - sign_keys True
+            - sign_keys False
+        - check_signature False
+    """
```
tomasr8: Rather than a class docstring, I'd prefer to put docstrings or comments into the individual methods. That way you have the comment where it's actually relevant and don't need to hunt for it.

KeyCloakAuthenticator/keycloakauthenticator/tests/test_auth.py
```diff
@@ -32,7 +34,239 @@ def _get_mock_token(private_key, token_id, expired=False):
...
+        {"authorization_endpoint": "http://fake/auth", "token_endpoint": "http://fake/token"},
+        {"token_endpoint": "http://fake/token"},
+    ])
+    async def test_missing_required_authorisation_fields(self, unconfigured_authenticator, monkeypatch, doc):
+        async def mock_httpfetch(url, **kwargs):
+            return doc
+
+        monkeypatch.setattr(unconfigured_authenticator, "httpfetch", mock_httpfetch)
+
+        with pytest.raises(Exception, match = "Unable to retrieve OIDC necessary values"):
```
tomasr8:
```suggestion
        with pytest.raises(Exception, match="Unable to retrieve OIDC necessary values"):
```

KeyCloakAuthenticator/keycloakauthenticator/tests/test_auth.py
```diff
@@ -32,7 +34,239 @@ def _get_mock_token(private_key, token_id, expired=False):
...
+            await unconfigured_authenticator._get_oidc_configs_helper()
+
+        assert unconfigured_authenticator.configured is False
```
tomasr8: Surprised ruff is not complaining about this
```suggestion
        assert not unconfigured_authenticator.configured
```

KeyCloakAuthenticator/keycloakauthenticator/tests/test_auth.py
```diff
@@ -32,7 +34,239 @@ def _get_mock_token(private_key, token_id, expired=False):
...
+            return OIDC_DISCOVERY_DOC
+
+        monkeypatch.setattr(unconfigured_authenticator, "httpfetch", mock_httpfetch)
+
+        await unconfigured_authenticator._get_oidc_configs_helper()
+
+        assert unconfigured_authenticator.authorize_url == "http://fake/auth"
+        assert unconfigured_authenticator.token_url == "http://fake/token"
+        assert unconfigured_authenticator.userdata_url == "http://fake/userinfo"
+        assert unconfigured_authenticator.configured is True
```
tomasr8:
```suggestion
        assert unconfigured_authenticator.configured
```

pyproject.toml
```diff
@@ -7,7 +7,6 @@ dev = [
     "watchfiles==1.2.0",
     "keycloakauthenticator @ file:///${PROJECT_ROOT}/KeyCloakAuthenticator",
     "swanculler @ file:///${PROJECT_ROOT}/SwanCuller",
-    "swanhub @ file:///${PROJECT_ROOT}/SwanHub",
```
tomasr8, line 10: Why was this removed?

tomasr8 (commented):
Overall looks good, I'd just check the line/branch coverage to make sure we're not missing anything (or conversely that we don't have any extra tests that are not needed)

tomasr8:
Also, it's better if you create separate branches for PRs rather than committing on your `master` :)

rodrigo-sobral (dismissed):

tomasr8 (approved):

PR #291, opened 2026-07-21
swanhub:  integration test for startup

icefox97:
integration test for the startup of swanhub with pre-populated user tables of various configs

### DIFF
--- a/SwanHub/tests/test_startup.py
+++ b/SwanHub/tests/test_startup.py
@@ -1,0 +1,192 @@
+import os
+import shutil
+import subprocess
+import textwrap
+import time
+from datetime import UTC, datetime
+from pathlib import Path
+
+import pytest
+import requests
+from jupyterhub import orm
+from sqlalchemy import create_engine
+from sqlalchemy.orm import Session
+
+HUB_URL = "http://127.0.0.1:18000"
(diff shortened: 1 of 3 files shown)


tomasr8:
This was the issue when swanhub was timing out I mentioned: https://github.com/swan-cern/jupyterhub-extensions/pull/225/changes
The problem was loading and parsing yaml for each spawner which is very CPU intensive which starved the event loop (when the hub starts, it creates a spawner instance for every user)

Could you re-apply that bug and see if your test catches it?

SwanHub/tests/test_startup.py
```diff
@@ -0,0 +1,192 @@
...
+
+import pytest
+import requests
+from jupyterhub import orm
+from sqlalchemy import create_engine
+from sqlalchemy.orm import Session
+
+HUB_URL = "http://127.0.0.1:18000"
+ADMIN_TOKEN = "test-admin-token"
+OPTIONS_FORM_CONFIG = Path(__file__).parents[2] / "SwanSpawner" / "options_form_config.json"
```
rodrigo-sobral: This `options_form_config.json` template should actually had been updated already, since we're now using yaml files.

rodrigo-sobral (approved):

PR #296, opened 2026-08-04
KeyCloakAuthenticator: Stop using preferred_username to identify users

rodrigo-sobral:
Per https://auth.docs.cern.ch/user-documentation/oidc/config/, it's not guaranteed to be unique. We should use either sub or cern_upn.

### DIFF
--- a/KeyCloakAuthenticator/README.md
+++ b/KeyCloakAuthenticator/README.md
@@ -29,7 +29,7 @@ In your JupyterHub config file, set the authenticator and configure it:
 ```python
 # Enable the authenticator
 c.JupyterHub.authenticator_class = 'keycloakauthenticator.KeyCloakAuthenticator'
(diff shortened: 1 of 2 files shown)


tomasr8:
Could you test with a secondary and a guest account just to see if there's any discrepancy when it comes to preferred_username and cern_upn/sub?

rodrigo-sobral:
> Could you test with a secondary and a guest account just to see if there's any discrepancy when it comes to preferred_username and cern_upn/sub?

Tested with my secondary account and I managed to start a session. Didn't work with a guest account since guests don't  have access to SWAN. Could you implement the same change on your charts and check if it works well for you too?

tomasr8 (approved):

tomasr8:
> Could you implement the same change on your charts and check if it works well for you too?

Checked and it works for me as well :+1:

PR #307, opened 2026-09-11
Default to 0 GPUs instead of raising when node are not ready in `AvailableGPUs`

### LINKED ISSUE #299: `AvailableGPUs` can sometimes raise when getting `Allocatable` info from a node
Talking about this line: https://github.com/swan-cern/jupyterhub-extensions/blob/def5a5cf800c752fdecc0862def694aecfc763a3/SwanSpawner/swanspawner/_gpuinfo.py#L319

I saw it raise when recreating some GPU nodes. I believe `AvailableGPUs` tried collecting info before the node was fully ready (operator done installing drivers, etc..).

This is rare enough, but still, we should default to zero here.

This might also be a great opportunity to create a basic test harness for `AvailableGPUs` (with the k8s bits mocked out).

thozo:
Addresses #299 

Created tests for the concerned function `_process_full_card`, as well as `_process_partitions` since the functions are similar. The `AvailableGPUs` mock is based on PR #297.

### DIFF
--- a/SwanSpawner/swanspawner/_gpuinfo.py
+++ b/SwanSpawner/swanspawner/_gpuinfo.py
@@ -316,7 +316,7 @@ class AvailableGPUs:
 
         # Check if the card is allocatable
         resource_name = 'nvidia.com/gpu'
-        count = int(node_status.allocatable[resource_name])
+        count = int(node_status.allocatable.get(resource_name, "0"))
         if count > 0:
             description = f'{gpu_model} ({memory} GB)'
             node_to_flavor[(node_name, resource_name)] = description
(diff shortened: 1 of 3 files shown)


SwanSpawner/swanspawner/_gpuinfo.py
```diff
@@ -316,7 +316,7 @@ def _process_full_card(self,
 
         # Check if the card is allocatable
         resource_name = 'nvidia.com/gpu'
-        count = int(node_status.allocatable[resource_name])
+        count = int(node_status.allocatable.get(resource_name, '0'))
```
tomasr8, line 319: `int(0) == 0` so we can simplify:

```suggestion
        count = int(node_status.allocatable.get(resource_name, 0))
```

PR #308, opened 2026-09-11
Remove dockerspawner from SwanSpawner

### LINKED ISSUE #305: Remove dockerspawner from swanspawner
It hasn't been used since we migrated to k8s and it's currently broken. We should just remove it instead of trying to fix it since it'll just bit rot anyway..

thozo:
Addresses #305


tomasr8 (approved):
Nice cleanup 👍

PR #312, opened 2026-09-15
SwanSpawner: Remove `SWANHUB_ENV` and kubespawner top level import

### LINKED ISSUE #306: Remove `SWANHUB_ENV` from `swanspawner`
This is a workaround to conditionally import the kubespawner only when used.

The kubespawner can only be imported inside a k8s cluster so for tests, etc., we currently set SWANHUB_ENV=true to get around this but that's not very nice.

We should simply **not** import the kube spawner in the top level swanspawner package at all.
That way you can import `swanspawner` without triggering an import of the kube spawner.
This will also need updating the `spawner_class` in the swan-charts repo.

thozo:
Addresses #306 

Remove all references to `SWANHUB_ENV` as the env var is now useless.

https://github.com/swan-cern/swan-charts/pull/589 should be merged before this PR.


tomasr8 (dismissed):
Nice cleanup!

PR #322, opened 2026-09-23
KeyCloakAuthenticator: fix check_signature

### LINKED ISSUE #92: Another bug in keycloakauthenticator - check_signarure is always true
Hello, team 
In keycloakauthenticator readme sample there are lines, that could disable signature verification, but this lines dont work, because check_signature is always True
https://github.com/swan-cern/jupyterhub-extensions/blob/0997d9bdd684bbdb977799b9a171d8d3ec5d337e/KeyCloakAuthenticator/README.md?plain=1#L66-L68
I think this behavior could be fixed if in lines
https://github.com/swan-cern/jupyterhub-extensions/blob/0997d9bdd684bbdb977799b9a171d8d3ec5d337e/KeyCloakAuthenticator/keycloakauthenticator/auth.py#L172-L173
and 
https://github.com/swan-cern/jupyterhub-extensions/blob/0997d9bdd684bbdb977799b9a171d8d3ec5d337e/KeyCloakAuthenticator/keycloakauthenticator/auth.py#L194-L195
you could change self.config.check_signature to self.check_signature

Could you please also fix this
or maybe I could make a PR?

thozo:
`self.config.check_signature` was checked, but the configured value was stored in `self.check_signature`.

Also added a log line when signature checks are disabled to notice potential misconfiguration.

Closes #92


tomasr8 (approved):
Nice, thanks!

